PR #7, opened 2018-11-20
added get_cop to heat_pump.py

Pyosch:
Initial discussion started [here](https://github.com/oemof/demandlib/issues/15)
Since the function is not in the `HeatBuilding ` class in demandlib.bdew anymore I changed the parameter `self ` to `heatbuilding`.


oemof_heat_components/heat_pump/heat_pump.py
```diff
@@ -0,0 +1,47 @@
...
+        and outside temperature
+    heatpump_type: string
+        defines the technology used. Ground is more efficient than Air.
+    water_temp: int
+        temperature needed for the heating system
+        
+    References
+    ----------
+    ..  [1]: 'https://www.researchgate.net/publication/255759857_A_review_of_domestic_heat_pumps'
+        Research paper about domestic heatpumps, containing the formulas used
```
jnnr: Add:
Returns
----------

jnnr:
Thanks for your effort! And sorry to see that this is quite a late reply. If this PR is still active, please do the following:
* Change base to dev.
* Place the module heat_pump.py in the new directory structure: src/oemof/thermal/heat_pump.py

Pyosch:
Thank you for your feedback! I hope it works for you now.

uvchik:
Thank you for your approach.

I do not know if it is a good idea to create a dependency on the demandlib. This will make it more difficult to use it for other purposes.

```python
def get_cop(building_temperature, heatpump_type="Air", water_temp=60):
...
return cop
```

If you want to use it together with the demandlib you can just use the following code:
```python
heatbuilding.cop = get_cop(heatbuilding.temperature)
```

For everybody else it would be much easier because users will not have to install the demandlib and instantiate a heat_building they do not want use just to pass a temperature time series.

uvchik:
@Pyosch We are planning to use the MIT licence in the future. Would you agree that your contribution will be under this licence? We need your agreement for the demandlib as well but you can just write a comment here for both contributions. Thanks in advance.

Pyosch:
@uvchik yes I don't mind. See you in Berlin ;)

src/oemof/thermal/heat_pump.py
```diff
@@ -0,0 +1,45 @@
...
+    elif heatpump_type == "Ground":
+        for tmp in temperature:
+            cop = (8.77 - 0.15 * (water_temp - tmp)
+                   + 0.000734 * (water_temp - tmp)**2)
+            cop_lst.append(cop)
+    
+    else:
+        traceback.print_exc("Heatpump type is not defined")
+        return None
+
```
uvchik, line 53: I do not want to be too petty and I think we can merge it without changing it but just for the protocol:

If an invalid parameter is given that leads to an invalid result you should raise an exception. If you return `None` the script will go on but we already know that it will lead to an exception later on. The user will have to scroll up to find the phrase "Heatpump type is not defined" between all the information given. As we know that `None` is not a valid COP we can raise a exception right away and the user can fix it directly. 
```diff
else:
-      traceback.print_exc("Heatpump type is not defined")
-      return None
+      msg = "'{0}' is not a valid heatpump type. Use 'Air' or 'Ground'."
+      raise ValueError(msg.format(heatpump_type)) 
```
The reason for such an error is most of the time just a typo. The user will get something like:
```
Value Error: 'AiR' is not a valid heatpump type. Use 'Air' or 'Ground'
```

uvchik (approved):

jnnr:
@jakob-wo, you have been looking into the heat pump topic recently. I guess your feedback would be really helpful here!

jnnr (approved):
Thanks for your effort! An example that shows the new function would be nice (in oemof-thermal/examples). You could get in contact with @jakob-wo, who is developing functions for heat pumps on a branch at the moment (not yet a PR: https://github.com/oemof/oemof-thermal/tree/features/cmpr_heatpumps_and_chillers/examples)

jakob-wo:
Hi @Pyosch,

thanks for enriching the COP calculation with field test data - I like that approach. 

As mentioned already above I am also about to come up with a COP calculation for compression heat pumps. I am planning to call the module `src/compression_heatpumps_and_chillers.py` [(features/cmpr_heatpumps_and_chillers)](https://github.com/oemof/oemof-thermal/tree/features/cmpr_heatpumps_and_chillers/src/oemof/thermal).
Furthermore, there is going to be a module on absorptions heat pumps somewhen in the near future.

It would be good to name your module clearly and distinguish between compression and absorption. I also think it might be helpful to make clear, that the module contains hard coded parameters/coefficients by including somethink like "field_test_regression" in the name.

jakob-wo:
One more remark:
The paper gives a temperature range the equations (Eq. 4) can be applied in.
Wouldn't it be nice to check whether the input data fullfills that condition and raise a UserWarning if someone exceeds the limit?

Pyosch:
@uvchik, @jnnr  and @jakob-wo thank you for your feedback! I'll look into it again. @jakob-wo your approach sounds very interesting too. I am looking forward to try it out.

jnnr:
Seems like travis does not run the tests properly. Might be because this PR comes from a fork. Anybody got an idea how to fix this?

Pyosch:
I added the warning messages and an example. The import of heat_pump inside the example is not correct I guess, but I couldn't quite figure out how the import is supposed to be structured. I looked at the other examples but it didn't help.

Any advice is appreciated :)

PR #8, opened 2019-07-25
Bring thermal components to oemof namespace

jnnr:
To make it easier to include the new components in oemof models, @gnn suggested to bring them to the same namespace. It works like this (adapted from oemof-tabular).

* [x] rename the repo: oemof-heat-components -> oemof-heat
* [ ] rename on rtd as well
* [x] change the directory structure: oemof-heat-components -> src/oemof/heat
* when added to Pypi, add as oemof.heat

This allows to import the new components like this:

`from oemof import heat`

A second suggestion: As we will cover technologies for cooling as well as heating, it would be more clear to call the package 'thermal' instead of 'heat', I guess. That would mean to replace 'heat' with 'thermal' in the structure above.


uvchik (approved):
I am glad that there is some activity in this direction. Go on :smile:

PR #14, opened 2019-10-16
Add files for testing and pylint / flake8 for style checks

jnnr:
Add a directory for tests.

Setup travis:
* pylint and flake8 for style checks
* pytest for autmated tests

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+exclude = meta/migrations/
+max-line-length=100
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 exclude = meta/migrations/
-max-line-length=100
+max-line-length=79
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,517 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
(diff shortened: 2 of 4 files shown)


.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+exclude = meta/migrations/
+max-line-length=100
```
uvchik: I know this is a matter of taste but in the other packages of oemof we use 79 according to PEP8.

.pylintrc
```diff
@@ -0,0 +1,517 @@
...
+
+# Force import order to recognize a module as part of a third party library.
+known-third-party=enchant
+
+
+[EXCEPTIONS]
+
+# Exceptions that will emit a warning when being caught. Defaults to
+# "Exception"
+overgeneral-exceptions=Exception
```
uvchik, line 517: I can not oversee the 500 lines of the `pylintrc` but the future will show how it works and we can adapt it to our needs later on.

.travis.yml
```diff
@@ -0,0 +1,50 @@
+language: python
+
+matrix:
+  include:
+    - python: 3.5
+    - python: 3.6
+    - python: 3.7
+      dist: xenial
+
```
uvchik, line 9: Most packages just drop python 3.5 support, as 3.6 is backward compatible we may drop it as well.

uvchik (commented):
Looks good. Just go ahead and try out :-)

PR #16, opened 2019-10-17
Add function for emission allocation for combined heat and power plants

jnnr:
The provided function allows to allocate emissions to the products of heat and power according to 3 different methods. See the source provided in the docstring for detailled information.


jakob-wo:
Useful bunch of calculations - thanks for the effort!

I have four comments on the PR:
1) I would rename `total_emissions` `total_specific_emissions` or `specific_emissions_fuel` because its unit is CO2/Energy. At least I recommend to change the docstring and clarify that you are dealing with specific emissions. The word "total" is misleading.
2) Fomulas (LaTeX) are missing in the docstring.
3) I would like to find another name for the module than `chp` because it is not very descriptive. Also, the module does not actually describe a technical unit (such as a chp) but legal regulation regarding the allocation of emissions. But till now I do not have any other suitable name... maybe `emissions_allocation_methods`.
4) flake8 is not happy yet :) some lines are to long.

Later I would like to add a little example. Maybe a plot that illustrates the differences between the methods, similar to the one in the paper mentioned in the references.

jnnr:
Thanks for your review!

>     1. I would rename `total_emissions` `total_specific_emissions` or `specific_emissions_fuel` because its unit is CO2/Energy. At least I recommend to change the docstring and clarify that you are dealing with specific emissions. The word "total" is misleading.

Thanks for pointing to that. Other than the reference, the provided function calculates the total (and not the specific) emissions. 

you can either write the formula for absolute emissions

EM_el =  A_Br * EM_gas

or specific emissions, which you then could multiply with the values of the model yourself:

specEM_el = A_Br * specEM_gas * W_gas / W_el

I decided for the former, as it seemed simpler to me. I added some units to the docstrings to make it easier to comprehend. Please tell me if you have an idea to improve it.

>     2. Fomulas (LaTeX) are missing in the docstring.

I added formulas.

>     3. I would like to find another name for the module than `chp` because it is not very descriptive. Also, the module does not actually describe a technical unit (such as a chp) but legal regulation regarding the allocation of emissions. But till now I do not have any other suitable name... maybe `emissions_allocation_methods`.

Good remark. I opened a related issue #36 on the location of post-processing functions. Indeed, this PR does not provide any other function for chp. We could create a new module called 'postprocessing', or 'emission_allocation'. The latter would however only contain a single function.

>     4. flake8 is not happy yet :) some lines are to long.

Travis does not throw any error. Where do you see flake8 complaining?

> Later I would like to add a little example. Maybe a plot that illustrates the differences between the methods, similar to the one in the paper mentioned in the references.

That would be useful. Maybe another plot would be even more interesting, like allocations as a function of efficiencies or something like that? Let's address this in a new issue / PR.

jakob-wo:
> Travis does not throw any error. Where do you see flake8 complaining?

I thought it did when I ran it locally but now I does not. Possible that I mixed something up. Sorry, my mistake.

src/oemof/thermal/chp.py
```diff
@@ -0,0 +1,93 @@
...
+        Electrical efficiency of the cogeneration [-].
+
+    eta_th : numeric
+        Thermal efficiency of the cogeneration [-].
+
+    method : str
+        Specification of method to use. Choose from ['iea', finnish', 'efficiency'].
+
+    **kwargs
+        For the finish method, `eta_el_ref` and `eta_th_ref` have to be passed.
```
jakob-wo:
```suggestion
        For the finnish method, `eta_el_ref` and `eta_th_ref` have to be passed.
```

jakob-wo: just a typo

src/oemof/thermal/chp.py
```diff
@@ -0,0 +1,93 @@
...
+
+    :math:`EM_{th} = EM \cdot \frac{\eta_{th}}{\eta_{el} + \eta_{th}}`
+
+    **Efficiency method**
+
+    :math:`EM_{el} = EM \cdot \frac{\eta_{th}}{\eta_{el} + \eta_{th}}`
+
+    :math:`EM_{th} = EM \cdot \frac{\eta_{el}}{\eta_{el} + \eta_{th}}`
+
+    **Finish method**
```
jakob-wo:
```suggestion
    **Finnish method**
```

jakob-wo: just a typo

jakob-wo (approved):
Looks good!

PR #18, opened 2019-10-31
Fix/docu api structure - reduce TOC depth level

jakob-wo:
This PR changes the structure of the API section in the documentation.

1) I wanted to reduce the depth-level of the TOC because in oemof-thermal we just have modules but no other packages. All modules shall be visible at the first level of the API overview.

2) I wanted to provide a structure we can use for the coming PRs of the components.


jnnr (approved):
Looks better! Thanks for fixing that.

jakob-wo:
Does anyone know what causes the checks to fail?

jnnr:
It fails because there are no tests yet in the test directory. We can ignore that for the moment.

uvchik:
You could also add a dummy test, because if test fail all the time won't notice if something else is wrong.

```python
def my_dummy_test()
    assert 5 == 5
```

FranziPl:
Looks good to me

FranziPl (commented):
Looks good to me.

FranziPl (approved):

jnnr:
> You could also add a dummy test, because if test fail all the time won't notice if something else is wrong.
> 
> ```python
> def my_dummy_test()
>     assert 5 == 5
> ```

Ok, I will add a dummy test.

uvchik:
> Make dummy test comparison non-redundant

Sorry @jnnr

jnnr:
No problem!

jakob-wo:
Great, thanks to both of you!

PR #19, opened 2019-11-06
Features/stratified thermal storage

jnnr:
Implementing the feature discussed in https://github.com/oemof/oemof/issues/621.

The example example_stratified_thermal_storage.py (without invest) is ready to be reviewed. The invest example is not finished yet.

Open tasks:
* [x] Merge https://github.com/oemof/oemof/pull/625 to implement the constant losses
* [x] Finish docs
    * [x] Finish text
    * [x] Finish table
    * [x] Write script that creates plots for the docs 
* [x] Finish storage investment example


MaGering:
The nominal storage capacity "Q_N" is missing as input parameter in docs/hot_water_storage.rst

symbol: `Q_N`
attribute: `nominal_storage_capacity`
explanation: Maximum amount of stored thermal energy

docs/hot_water_storage.rst
```diff
@@ -2,4 +2,168 @@
...
+
+  \gamma &= UA \Delta T_{H0} \Delta t.
+
+
+
+In the case of investment, the diameter d is given and the height can be
+adapted to adapt the nominal capacity of the storage. With this assumption,
+all relations stay linear.
+
+These parameters are input to the component:
```
MaGering: "input and output"?

Table contains output parameters:

- beta
- gamma
- eta_in
- eta_out

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,174 @@
...
+    return u_value
+
+
+def calculate_storage_dimensions(height, diameter):
+    r"""
+    Calculates volume and total surface of a hot water storage.
+
+    :math:`V = \pi \frac{d^2}{4} \cdot h`
+
+    :math:`A = \pi d h + 2 \pi \frac{d^2}{4}`
```
MaGering: For reasons of clarity, I would suggest: `A = \pi d h + \pi \frac{d^2}{2}`

Same applies for 

- line 73 (surface = np.pi * diameter * height + np.pi * diameter**2 * 1 / 2) and
- docs/hot_water_storage.rst line 164

docs/hot_water_storage.rst
```diff
@@ -2,4 +2,168 @@
...
+
+    :math:`d`                 :py:obj:`diameter`                                        Diameter
+
+    :math:`A`                 :py:obj:`surface`                                         Storage surface
+
+    :math:`V`                 :py:obj:`volume`                                          Storage volume
+
+    :math:`v_{nonuseable}`    :py:obj:`nonusable_storage_volume`                        Factor describing non-usable
+                                                                                        storage volume
+                                                                                        (:math:`0<v_{nonusable}<1`)
```
MaGering: If you add a new line here "(0<v_{nonusable}<1)" will not project beyond the table anymore

examples/stratified_thermal_storage/example_stratified_thermal_storage.py
```diff
@@ -0,0 +1,170 @@
...
+    print(dash)
+
+
+print_results()
+
+# Set up an energy system model
+solver = 'cbc'
+periods = 100
+datetimeindex = pd.date_range('1/1/2019', periods=periods, freq='H')
+x = np.arange(periods)
```
MaGering: Is x used somewhere else? What is it initialized for?

examples/stratified_thermal_storage/example_stratified_thermal_storage.py
```diff
@@ -0,0 +1,170 @@
...
+        'Max. heat flow charging [MW]': maximum_heat_flow_charging,
+        'Max. heat flow discharging [MW]': maximum_heat_flow_discharging,
+        'Max storage level [-]': max_storage_level,
+        'Min storage_level [-]': min_storage_level,
+        'Loss rate [-]': loss_rate,
+        'Fixed relative losses [-]': fixed_losses_relative,
+        'Fixed absolute losses [MWh]': fixed_losses_absolute,
+    }
+
+    dash = '-' * 42
```
MaGering: Suggestion for improvement of screen layout: 50 dashes (see example_stratified_thermal_storage_investment.py).

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+        Total surface of storage [m2]
+    """
+    volume = diameter**2 * 1 / 4 * np.pi * height
+    surface = np.pi * diameter * height + 2 * np.pi * diameter**2 * 1 / 4
+
+    return volume, surface
+
+
+def calculate_capacities(volume, temp_h, temp_c, nonusable_storage_volume,
+                         heat_capacity=4180, density=971.78):
```
MaGering: heat_capacity=4180 -> water at approx. 30 °C
density=971.78 -> water at approx. 80 °C

How were those parameters chosen?

examples/stratified_thermal_storage/example_stratified_thermal_storage.py
jnnr: You are right, it is obsolete.

src/oemof/thermal/stratified_thermal_storage.py
jnnr: Good point! Actually, this function is an approximation, as it assumes constant density and heat capacity independent of temperature. To be more correct, we could write a more elaborate function using coolprop. Coolprop might be useful for other components as well. If we want to start with this simple approach, we need some assumption on the numbers. What do you think @FranziPl?

jnnr:
Can you please review before the end of the week @FranziPl? Just want to make sure that we are able to release on Monday.

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    volume = diameter**2 * 1 / 4 * np.pi * height
+    surface = np.pi * diameter * height + np.pi * diameter**2 * 0.5
+
+    return volume, surface
+
+
+def calculate_capacities(volume, temp_h, temp_c, nonusable_storage_volume,
+                         heat_capacity=4180, density=971.78):
+    r"""
+    Calculates the nominal storage capacity, surface area, minimum
```
FranziPl: I don't see, where it calculates the surface area?

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    and maximum storage level of a stratified thermal storage.
+
+    .. calculate_capacities-equations:
+
+    :math:`Q_N = \pi \frac{d^2}{4} \cdot h \cdot c \cdot \rho \cdot \left( T_{H} - T_{C} \right)`
```
FranziPl: You have to give the volume as an input, not height and diameter. So I would adjust the formula.

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+
+    :math:`Q_{max} = Q_N \cdot (1-v_{nonusable}/2)`
+
+    :math:`Q_{min} = Q_N \cdot v_{nonusable}/2`
+
+    Parameters
+    ----------
+    volume :numeric
+        Volume of the storage[m]
```
FranziPl: volume unit should be [m^3]

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+
+    max_storage_level : numeric
+        Maximal storage content relative to nominal storage capacity [-]
+
+    min_storage_level : numeric
+        Minimal storage content relative to nominal storage capacity [-]
+
+    """
+    nominal_storage_capacity = 1 / 3600 * volume * heat_capacity * density * (temp_h - temp_c)
+    nominal_storage_capacity *= 1e-6  # [Wh] to [MWh]
```
FranziPl: If you have a single step to change the unit from Wh to MWh, maybe there also should be a single step to change unit from J to Wh.

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    max_storage_level = (1 - nonusable_storage_volume / 2)
```
FranziPl: Why are the brackets here? Is it a style guide thing?

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    Returns
+    -------
+    volume : numeric
+        Volume of storage
+
+    surface : numeric
+        Total surface of storage [m2]
+    """
+    volume = diameter**2 * 1 / 4 * np.pi * height
+    surface = np.pi * diameter * height + np.pi * diameter**2 * 0.5
```
FranziPl: Just for readability, I would make the two formulas look similar:
surface = np.pi * diameter * height + diameter**2 * 1 / 2 * np.pi

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    height : numeric
+        Height of the storage [m]
+
+    diameter : numeric
+        Diameter of the storage [m]
+
+    Returns
+    -------
+    volume : numeric
+        Volume of storage
```
FranziPl, line 69: Volume of storage [m3]

src/oemof/thermal/stratified_thermal_storage.py
FranziPl: I am not sure about that. You use this xalues also in the "calculate_losses"-function. And if you want to use the correct values, there you would have to use different values for the cold and the hot zone, which whould make your formulas more complicated. I think one of the simplifications for the approach are constant values for c and rho. So you could coolprop for the average temperature. As this is also a simplification and causes an error, I think it is not necessary to write a coolprop function. I would just adapt the values to water whith 80 degrees. (Was that understandable? If not, please call me :))

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    fixed_losses_relative : numeric (sequence or scalar)
+        Losses independent of state of charge between two consecutive
+        timesteps relative to nominal storage capacity [-]
+
+    fixed_losses_absolute : numeric (sequence or scalar)
+        Losses independent of state of charge and independent of
+        nominal storage capacity between two consecutive timesteps [MWh]
+    """
+    loss_rate = (
+        4 * u_value * (temp_h - temp_c) * 1 / (diameter * density * heat_capacity) * time_increment
```
FranziPl: 4 * u_value / (diameter * density * heat_capacity) * time_increment

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+    )
+
+    fixed_losses_relative = (
+        4 * u_value * 1 / (diameter * density * heat_capacity)
```
FranziPl: 4 * u_value / (diameter * density * heat_capacity * (temp_h - temp_c))

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -0,0 +1,204 @@
...
+        * (temp_c - temp_env) * time_increment
+    )
+
+    fixed_losses_absolute = (
+        np.pi * diameter**2 * 0.25 * (temp_h + temp_c - 2 * temp_env) * time_increment
```
FranziPl: u_value * np.pi * diameter**2 / 4 * (temp_h + temp_c - 2 * temp_env) * time_increment

FranziPl (commented):
I didn't review the examples yet, it is just for the stratifies_thermal_storage.py.
I left comments in the code, but I also have 3 general points:
It is common, to have the output unit in MWh? I thought, the more common one is kWh?
I was told, that there should be period in the end of a sentence in the docstrings (for example:
s_iso : numeric
        Thickness of isolation layer [mm].     )Is this correct?
The formulas in the function calculate losses are differnt to the formulas in the docstrings. I assumed that the ones in the docstring are correct. Is this correct?
I will check the other documents later.

docs/stratified_thermal_storage.rst
```diff
@@ -0,0 +1,196 @@
...
+  for outside temperatures for the calculation of heat losses.
+* There is no distinction between outside temperature and ground temperature.
+* Material properties are constant.
+
+The equation describing the change of storage content is the following:
+
+.. math::
+  Q_t = Q_{t-1} \Big(1- U \frac{4}{D\rho c}\Delta t\Big)
+  - U \frac{4Q_N}{D\rho c \Delta T_{HC}}\Delta T_{C0}\Delta t
+  - U \frac{\pi D^2}{4}\Big(\Delta T_{H0} + \Delta T_{C0}\Big)
```
FranziPl: I think there is a \Delta t missing?

docs/stratified_thermal_storage.rst
```diff
@@ -0,0 +1,196 @@
...
+with
+
+.. math::
+  \beta &= U \frac{4}{D\rho c}\Delta t
+
+  \gamma &= U \frac{4}{D\rho c \Delta T_{HC}}\Delta T_{C0}\Delta t
+
+  \delta &= U \frac{\pi D^2}{4}\Big(\Delta T_{H0} + \Delta T_{C0}\Big) \Delta t.
+
+
```
FranziPl, line 55: I think it whould be helpful to explain, which losses this terms are representing in real physics (Losses from bottom and roof of the storage, losses from hot side walls, losses from cols side walls).

docs/stratified_thermal_storage.rst
```diff
@@ -0,0 +1,196 @@
...
+    :math:`\rho`              :py:obj:`density`                          Density of storage
+                                                                         medium
+
+    :math:`c`                 :py:obj:`heat_capacity`                    Heat capacity of
+                                                                         storage medium
+
+    :math:`T_H`               :py:obj:`temp_h`                           Hot temperature level
+
+    :math:`T_C`               :py:obj:`temp_c`                           Cold temperature
```
FranziPl: :math:`T_C`               :py:obj:`temp_c`                           Cold temperature level

docs/stratified_thermal_storage.rst
```diff
@@ -0,0 +1,196 @@
...
+
+    :math:`\beta`             :py:obj:`loss_rate`                        Relative loss of
+                                                                         storage content
+                                                                         within one timestep
+
+    :math:`\gamma`            :py:obj:`fixed_losses`                     Fixed losses as share
+                                                                         of nominal storage
+                                                                         capacity
+
+    :math:`\delta`            :py:obj:`fixed_absolute_losses`            Fixed absolue losses
```
FranziPl: ... Fixed absolute losses

examples/stratified_thermal_storage/stratified_thermal_storage.csv
```diff
@@ -0,0 +1,15 @@
...
+density,971.78,kg/m3
+heat_capacity,4180,J/(kg*K)
+temp_h,95,degC
+temp_c,60,degC
+temp_env,10,degC
+inflow_conversion_factor,0.9,
+outflow_conversion_factor,0.9,
+nonusable_storage_volume,0.05,
+s_iso,0.05,m
+lamb_iso,0.03,W/(m*K)
```
FranziPl, line 13: Did you check, if these 3 values are senseful? I think, at least one of the alphas is too small. The resulting plot has almost no losses depending on the storage content. I think this way the plot is misleading and doesn't show the difference between this aproach and the simple storage appropriately.

examples/stratified_thermal_storage/plots.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+ax2.set_xlim(0, 3.2)
+ax2.set_ylim(0, 0.0035)
+ax2.set_title('Losses vs. storage content')
+ax2.set_xlabel('Storage content [MWh]')
+ax2.set_ylabel('Losses [MW]')
+ax2.grid(alpha=0.3)
+ax2.legend(loc='center left', bbox_to_anchor=(1.0, 0.5))
+plt.tight_layout()
+plt.savefig('compare_storage_models.svg')
```
FranziPl, line 191: The picture, I created with this example, looks different to the picture in the documentation. Did you change parameters, or what is the difference?

FranziPl:
I'm done with the review.

src/oemof/thermal/stratified_thermal_storage.py
jnnr: Change the values to those for 80°C. We should discuss about using coolprop generally at a later point in time. I guess this question comes for other components as well.

examples/stratified_thermal_storage/stratified_thermal_storage.csv
jnnr: There was an error in the loss calculation, as the heat capacity was not converted from J to Wh. This lead to very small numbers for beta and gamma.

examples/stratified_thermal_storage/plots.py
jnnr: Might be because you have not checked out the correct oemof branch that has the fixed_losses attributes. oemof currently does not warn you when you pass non-existing keyword arguments to a component. Make sure you have v0.3 checked out.

jnnr:
> I'm done with the review.

Thanks for your effort and the useful remarks, @FranziPl! I adapted the code and the examples accordingly. Please tell me if there are other issues.

FranziPl (approved):
I am happy with it now.

PR #26, opened 2019-12-10
Add list with examples to documentation

jakob-wo:
This PR adds a section to the docu containing a list with available examples.


FranziPl (approved):
I like the section and think it helps, to make the documentation more useful.
In a next step, I whould add a sentence for each example to explain, what it does.

Issue #23, opened 2019-11-14
Release of v0.0.1


c-moeller:
We set a milestone for the first release of oemof-thermal by December 24th.

@oemof/oemof-heat Let's use this issue to discuss what need to be done to achieve this.

jnnr:
Thanks for bringing this up. The release should contain the components that are in review at the moment.

#7
#16
#19
#20
#21
#22

What is the status of #7?

For the actual release process, we can consult https://github.com/oemof/oemof/wiki/How-to-release

uvchik:
> December 24th

:joy: :santa: :christmas_tree:

jnnr:
I adapted the date for the release to a more realistic one (16/12/2019). Still, we can regard it as a christmas present to ourselves :)

To be able to present the new features at the upcoming dev meeting, we aim to finish reviews and merge into dev until 04/12/2019.

c-moeller:
* [x] Remove png files from CSP and flat plate collector example
* [x] Remove lp file from stratified storage example
* [x] Correct plot in CSP example (there is an empty plot)

jnnr:
* [x] install_requires are not installed
* [x] Example overview lacks flat plate collector
* [x] inconsistant naming: Example is called flat plate collector, module solar thermal collector

c-moeller:
What does the footnote "1" mean in the examples doc?

FranziPl:
> 
> 
> What does the footnote "1" mean in the examples doc?

[1] Example uses *oemof.solph*


(I have no idea, why it is build in this strange way)

jnnr:
> (I have no idea, why it is build in this strange way)

The numbers 1,2,3,4,5 are the links back from where the footnote came.  Might be useful for a large document. In this case this feature is rather confusing.

c-moeller:
We are now just before merging the release branch into the master branch. I will start with this tomorrow and complete the release.

c-moeller:
oemof.thermal is now on Pypi and the release (almost) completed:
https://pypi.org/project/oemof.thermal/0.0.1/

Does anyone know if the documentation is built automatically? Till now there is no v0.0.1 at readthedocs: https://oemof-thermal.readthedocs.io/en/latest/

There are some few points I would like to do in the beginning of January:
* [x] Wordpress news post --> I would like to do that in the new year to avoid that the news disappear due to christmas holidays
* [x] Edit Whatnew file --> together with the news post
[...]

PR #32, opened 2019-12-17
Releases/v0.0.1

jnnr:
This is the first release of oemof-thermal. Relates to #23.


jnnr:
* [ ] install_requires are not installed
* [ ] Example overview lacks flat plate collector

Issue #34, opened 2020-01-06
There are no tests


jnnr:
We do not have any tests yet. It would be very helpful for developers to have some automatic tests. I suggest that we do the following: First, think about how reasonable tests for the package would look like. Then, the contributors to the different modules write some tests for the functions.

TODO

* [x] Write tests for solar_thermal_collector

jnnr:
There are PRs underway that introduce tests. Leave this issue open until a test coverage metric is introduced.

jakob-wo:
What about this Issue? Can we close it?

jnnr:
There are no tests yet for the solar thermal collector.

jakob-wo:
Of course, you're right! I somehow missed that package - sorry!

jnnr:
No problem! Thanks for asking :)

jnnr:
To get a detailed view of the parts of the code that are covered and which not, you can use the following command to generate an report in html form

`pytest --cov --cov-report html`

PR #38, opened 2020-01-06
Add tests for storage preprocessing functions

jnnr:
Adds tests for the stratified thermal storage module as a first sketch of possible tests. Adresses #34.


FranziPl:
@jnnr Sorry, I didn't realize I was the reviewer earlier.
Should I check the values too, or is this just about test structure?

jnnr:
@FranziPl No worries.

I checked the values, so they should be ok. But if you want to check the values again and (!) it does not take you too much time, go for it.

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'diameter': 10,  # m
+    }
+
+    volume, surface = calculate_storage_dimensions(**params)
+    assert volume == approx(250 * np.pi) and surface == approx(150 * np.pi)
+
+
+def test_calculate_capacities():
+    params = {
+        'volume': 1000,
```
FranziPl:
```suggestion
        'volume': 1000,  # m3
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'temp_h': 100,
```
FranziPl:
```suggestion
        'temp_h': 100,  # deg C
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'temp_c': 50,
```
FranziPl:
```suggestion
        'temp_c': 50,  # deg C
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+
+    nominal_storage_capacity, max_storage_level, min_storage_level = calculate_capacities(**params)
+    assert nominal_storage_capacity == 56.62268888888889\
+        and max_storage_level == 0.95\
+        and min_storage_level == 0.05
+
+
+def test_calculate_losses():
+    params = {
+        'u_value': 1,
```
FranziPl:
```suggestion
        'u_value': 1,  # W/(m2*K)
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'diameter': 10,
```
FranziPl:
```suggestion
        'diameter': 10,  # m
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'temp_h': 100,
```
FranziPl:
```suggestion
        'temp_h': 100,  # deg C
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'temp_c': 50,
```
FranziPl:
```suggestion
        'temp_c': 50,  # deg C
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'temp_env': 10,
```
FranziPl:
```suggestion
        'temp_env': 10,  # deg C
```

tests/test_functions.py
```diff
@@ -1,3 +1,59 @@
...
+        'time_increment': 1
```
FranziPl: As the time_increment has the default value 1, I would omit it.

FranziPl (approved):
I wanted to calculate the values too, that's why I waited so long, but in the end I didn't. So I just checked the structur, but not the asserted values.
I have some minor suggestions: to make the form more consistent, I would add the units everywhere.

tests/test_functions.py
jnnr: It is unnecessary to specify this, you are right.

jnnr:
Thanks, @FranziPl!

I tried to use numbers that are easy to check. I think it is good to cross-check the values asserted in these tests, because once they are set they will hopefully not be touched for a while, so it is good to rely on them.

@MaGering: Can you please have a quick look and check if the values in these tests are fine? It should not take too long, I guess.

MaGering:
I've reviewed the tests. They were all correct except for test_calculate_capacities() and test_calculate_losses().

This is due to 

- deviant default values for the density and capacity, see: https://github.com/oemof/oemof-thermal/pull/70 and

- the nominal_storage_capacity missing in the multiplication of the max_storage_level and min_storage_level, see: https://github.com/oemof/oemof-thermal/pull/71

I will write the corrected test results of those two tests inside a comment in the files changed section.

tests/test_functions.py
```diff
@@ -1,9 +1,62 @@
...
+def test_calculate_capacities():
+    params = {
+        'volume': 1000,  # m3
+        'temp_h': 100,  # deg C
+        'temp_c': 50,  # deg C
+        'nonusable_storage_volume': 0.1,  # dimensionless
+    }
+
+    nominal_storage_capacity, max_storage_level, min_storage_level = calculate_capacities(**params)
+    assert nominal_storage_capacity == 56.62268888888889\
```
MaGering: assert nominal_storage_capacity == 56.62804059111111\

tests/test_functions.py
```diff
@@ -1,9 +1,62 @@
...
+        and max_storage_level == 0.95\
+        and min_storage_level == 0.05
```
MaGering, line 44: and max_storage_level == 53.796638561555554\
        and min_storage_level == 2.8314020295555555

tests/test_functions.py
```diff
@@ -1,9 +1,62 @@
...
+        'temp_env': 10,  # deg C
+    }
 
-def test_dummy():
-    a = 2 * 0.5
-    assert a == 1
+    loss_rate, fixed_losses_relative, fixed_losses_absolute = calculate_losses(**params)
+    assert loss_rate == 0.00035450164405061065\
+        and fixed_losses_relative == 0.00028360131524048847\
+        and fixed_losses_absolute == 0.010210176124166827
```
MaGering: assert loss_rate == 0.0003531819182021882\
        and fixed_losses_relative == 0.00028254553456175054\
        and fixed_losses_absolute == 0.010210176124166827

jnnr:
>     * deviant default values for the density and capacity, see: #70 and

Before adapting this, we should finalize #70, merge it to dev and then merge dev into this branch.

>     * the nominal_storage_capacity missing in the multiplication of the max_storage_level and min_storage_level, see: #71

I don't think this is necessary. I commented on that in #71.

tests/test_functions.py
jnnr: This is not necessary. The values are correct. They are relative to the nominal_storage_capacity.

tests/test_functions.py
jnnr: Adapted this.

tests/test_functions.py
jnnr: Adapted this.

Issue #35, opened 2020-01-06
There are no facades for components in oemof-thermal yet


jnnr:
At the past dev meeting in Berlin, we had a session on facades in oemof. The session is documented here: https://github.com/oemof/organisation/wiki/Meeting-December-2019#facades and the decisions are summarized in this issue: https://github.com/oemof/oemof/issues/594

The proposals that emerged from the session could be implemented as a test. If we like it, we could establish it in the next release.

PR #43, opened 2020-01-08
Features/facades

jnnr:
This PR implements a facade `StratifiedThermalStorage`. Related issue: #35 
It basically wraps some of the precalculations into a new class, which makes it easier to parametrize the component.

**How to use the new facade class**
To test it, first run the existing examples without facades
* example_stratified_thermal_storage.py
* example_stratified_thermal_storage_investment.py

and then the new ones using facades
* example_stratified_thermal_storage_facades.py
* example_stratified_thermal_storage_investment_facade.py

The examples produce lp-files, and the latter ones check if both are the same. This is a dummy to test the concepts we have been discussing in practice, so any comments are welcome.

Make sure that you use the dev version of oemof: https://github.com/oemof/oemof/tree/dev

The facades are derived from oemof.tabular. I refrained from importing oemof.tabular, as it comes with many additional dependencies, and just copied the baseclase Facade from the respective module. 

**Open todos and solved questions**:

* [x] Allow to pass precalculation functions with the given functions as default.
   This refers to `calculate_losses`, `calculate_capacities` and `calculate_storage_dimensions`. In principle it would be easy to set the functions as default arguments and make it possible for the user to replace it. The question is how probable it is that someone wants to do this. In my view, if someone is already at that level, they might want to look inside the facade class theirself. Maybe later or for other facades we might think differently, but I would leave it like this for here and now.
* [x] Should it be possible to pass heat_capacity and density?
   Yes. Adapted it so the parameters are passed to calculate_capacities and calculate_losses if they are given.
* [x] Play around with the facades to check if they behave nicely.
* [x] Make sure that results handling correct (outputlib)
* [x] Implement the investment example.
* [x] Discuss if investment in energy with fixed power should be possible (opened issue here #63). Update: If nobody objects, this can be postponed.
   * [x] Invest into nominal_storage_capacity but leaving charging/discharging power fixed
    Seems to be rare or not relevant.
   * [x] Invest into nominal_storage_capacity and charging/discharging power with a fixed ratio: This works already
   * [x] Invest into nominal_storage_capacity and charging/discharging power independently without a fixed ratio: This works already
* [x] Move the constraint tests from src/oemof/thermal to ./tests and 
* [x] setup some more tests.

Comments moved from below:

* [x] Default arguments are not set consitently, e.g. diameter=0 etc.
       Diameter is now required and no defaults are set, apart from capacity in investment.
* [x] Unsetting default zero for capacity breaks the investment 
       I changed it so only in investment mode, capacity is zero by default.
* [x] facade_requires should include all required arguments for the storage
* [x] in build_solph_components: is it necessary to call sequences(...) in all cases, or isn't that already happening within the oemof-solph component?
   Yes, this is necessary. The same happens in the __init__ method of the components in solph.
* [x] self._investment() returns an investment object that is not used.
   Actually, that's not true. The return of self._investment() is assigned to self.investment. This is the investment object of the storage. If the storage's charging/discharging power is invested as well, investment objects will be passed to the Flows. 
* [x] carrier and tech are tabular requirements that cannot be understood if you don't have that context.
  I just removed it from the facade requires, the examples and the tests. To run it with oemof.tabular, just add them again to the facade_requires.

A separate question:
* [x] Decide if we want facades for the other 3 existing components of oemof.thermal.
   Outcome: Solar thermal collector and csp could benefit. Not that pressing for heat pump, but good for completeness. Can be implemented after the solution has proved practicable with the other components.


jnnr:
@FranziPl: Please have a look at the two new examples using facades. Try if they work on your system. You don't need to go too deep into the implementation behind. But some feedback on the interface and usage of the facade would be helpful.

jnnr:
* [x] Update: When handling the results with the outputlib there seems to be a problem. The StratifiedThermalStorage does not appear in the dict keys. Maybe I just overlooked something, but this has to be checked.

I checked that again but could not reproduce the error. Both the original examples as the ones using facades give the same results.

FranziPl:
I moved these comments to the top.

jnnr:
@MaGering: If you find the time, can you try out the new facade class StratifiedThermalStorage? I did not write the docs yet, but I hope the information in this PR is enough to understand it. If not, just let me know/call me :)

jnnr:
@FranziPl: I adapted the points that we noted some weeks ago. There are only a few things open (see above). If you find the time, could you have another review here?

@jakob-wo: You will use the component in validation I guess, so if you have any feedback on this, I am happy to read about it :)

src/oemof/thermal/facades.py
```diff
@@ -0,0 +1,330 @@
...
+
+        self.height = kwargs.get("height")
+
+        self.temp_h = kwargs.get("temp_h")
+
+        self.temp_c = kwargs.get("temp_c")
+
+        self.temp_env = kwargs.get("temp_env")
+
+        self.u_value = kwargs.get("u_value")
```
MaGering:
```suggestion
        # self.diameter = kwargs.get("diameter")

        self.height = kwargs.get("height")

        # self.temp_h = kwargs.get("temp_h")

        # self.temp_c = kwargs.get("temp_c")

        # self.temp_env = kwargs.get("temp_env")

        # self.u_value = kwargs.get("u_value")
```

No error message without commented out values of my proposition, because of:
> super().__init__(
>              _facade_requires_=[
>                  "bus", "diameter",
>                  "temp_h", "temp_c", "temp_env",
>                  "u_value"], *args, **kwargs

Are these values redundant?

src/oemof/thermal/facades.py
```diff
@@ -0,0 +1,330 @@
...
+
+        self.water_properties = {
+            'heat_capacity': kwargs.get("heat_capacity"), 'density': kwargs.get("density")
+        }
+
+        self.capacity = kwargs.get("capacity")
+
+        self.storage_capacity_cost = kwargs.get("storage_capacity_cost")
```
MaGering:
```suggestion
        # self.storage_capacity_cost = kwargs.get("storage_capacity_cost")
```

There is no consequence if this argument is missing.

src/oemof/thermal/facades.py
jnnr: You are right. Arguments that are in the facade_requires are set in the __init__() of the parent class. That means that they do not have to be set again.

src/oemof/thermal/facades.py
```diff
@@ -0,0 +1,330 @@
...
+            "storage_capacity_potential", float("+inf")
+        )
+
+        self.capacity_potential = kwargs.get(
+            "capacity_potential", float("+inf")
+        )
+
+        self.minimum_storage_capacity = kwargs.get(
+            "minimum_storage_capacity", 0
+        )
```
MaGering:
```suggestion
        # self.capacity_cost = kwargs.get("capacity_cost")

        # self.storage_capacity_potential = kwargs.get(
           #  "storage_capacity_potential", float("+inf")
        # )

        # self.capacity_potential = kwargs.get(
            # "capacity_potential", float("+inf")
       #  )

        # self.minimum_storage_capacity = kwargs.get(
            # "minimum_storage_capacity", 0
        # )
```

Just as with 'storage_capacity_cost', there are no consequences (no error message) if the following arguments are missing:

- capacity_cost
- storage_capacity_potential
- capacity_potential
- minimum_storage_capacity

cf. commented out code in my proposition

src/oemof/thermal/facades.py
```diff
@@ -0,0 +1,330 @@
...
+        self.fixed_losses_relative = losses[1]
+
+        self.fixed_losses_absolute = losses[2]
+
+        self.build_solph_components()
+
+    def build_solph_components(self):
+        """
+        """
+        self.inflow_conversion_factor = sequence(self.efficiency)
```
MaGering, line 258:
```suggestion
        # self.inflow_conversion_factor = sequence(self.efficiency)
```

There is no consequence if this argument is missing.

MaGering:
I've tested the input arguments of thermal_storage = facades.StratifiedThermalStorage(...):

- diameter
- height
- temp_h
- temp_c
- temp_env
- u_value
- heat_capacity
- density
- capacity
- storage_capacity_cost
- capacity_cost
- storage_capacity_potential
- capacity_potential
- minimum_storage_capacity
- expandable
- efficiency
- marginal_cost
- input_parameters
- output_parameters

These arguments were derived from StratifiedThermalStorage class in facades.py. I alternated their values and checked the lp files against each other. They mostly behaved nicely! I've gathered arguments I considered striking in two example files, specially created for my argument testing:

https://github.com/oemof/oemof-thermal/pull/43/commits/5d5ce87c2794db1e07148ed2e92021d949177342

What struck me is:

- As a user of facades I would not be so sure about what arguments I can pass
- I can pass nonsense values, that are not considered in further calculations. This might cause some confusion in my opinion, because I wouldn't know as a user, if the arguments I passed were considered in the facades or not (but I might be wrong here since I'm not that experienced with facades)
- Related to the before mentioned points: I was confused, wether I can pass further values from 'stratified_thermal_storage.csv' or not to the facades before I checked the class in facades.py
- The variation of temp_env does not seem to have an influence on the comparison of the lp files (even though the losses are calculated from it inside facades.py)

src/oemof/thermal/facades.py
jnnr: I guess this is because in GenericStorage, inflow_conversion_factor has a default value of 1, which is the same value that was assumed in the example.

jnnr:
> What struck me is:
> 
>     * As a user of facades I would not be so sure about what arguments I can pass

This should be improved by adding documentation. Have a look at #98 if this is resolved or needs to be extended.

>     * I can pass nonsense values, that are not considered in further calculations. This might cause some confusion in my opinion, because I wouldn't know as a user, if the arguments I passed were considered in the facades or not (but I might be wrong here since I'm not that experienced with facades)

Yes, at the moment there is no such check. I would suggest to open a new issue to discuss if this would make sense.

>     * Related to the before mentioned points: I was confused, wether I can pass further values from 'stratified_thermal_storage.csv' or not to the facades before I checked the class in facades.py

The documentation, the examples and the docstrings should help. Got suggestions for improvement?

>     * The variation of temp_env does not seem to have an influence on the comparison of the lp files (even though the losses are calculated from it inside facades.py)

I wrote comments next to the code suggestions and the two examples. I deleted the redundant setting of kwargs that are in facade_requires. I removed the calculation of min_storage_level/max_storage_level from nonusable_storage_volume, as it is trivial. In #98, we should add a sentence on the idea of a part of the volume being unusable.

If you are fine with that, we can delete the two examples that you added.

examples/stratified_thermal_storage/example_stratified_thermal_storage_facade_arg_test.py
```diff
@@ -0,0 +1,122 @@
...
+    temp_h=input_data['temp_h'],
+    temp_c=input_data['temp_c'],
+    u_value=u_value,
+    min_storage_level=min_storage_level,
+    max_storage_level=max_storage_level,
+    capacity=maximum_heat_flow_charging,
+    efficiency=1,
+    marginal_cost=0.0001,
+    # Arguments that do not play a role in comparison of lp files
+    temp_env=1000,
```
jnnr: I tested it: It makes a difference when you use the correct oemof branch. If not, the GenericStorage misses the two extra fixed loss terms.

examples/stratified_thermal_storage/example_stratified_thermal_storage_facade_arg_test.py
```diff
@@ -0,0 +1,122 @@
...
+    nonusable_storage_volume=1000,  # Argument not integrated in facades.py but in stratified_thermal_storage.csv
```
jnnr: This is the same as banana: You can pass it, but it is not used.

examples/stratified_thermal_storage/example_stratified_thermal_storage_investment_facade_arg_test.py
```diff
@@ -0,0 +1,103 @@
...
+    min_storage_level=0.025,
+    max_storage_level=0.975,
+    efficiency=1,
+    marginal_cost=0.0001,
+    # Arguments that do not play a role in comparison of lp files
+    temp_env=1000,
+    nonusable_storage_volume=1000,  # Argument not integrated in facades.py but in stratified_thermal_storage.csv
+    # Arguments that do not have default values set yet
+    capacity_potential=float("+inf"),
+    storage_capacity_potential=float("+inf"),
```
jnnr: That is not true. The default values are set in
https://github.com/oemof/oemof-thermal/blob/b9e6f3b62e756e8ddc4270926bd4759682d8778c/src/oemof/thermal/facades.py#L223-L229

MaGering: Yes, that's true. What I meant (but formulated poorly) is that they are not listed in the example of the investment facade so far. Might it be of interest for the user to have it listed in the example?

jnnr: Or maybe in the docs?

MaGering:
I've fixed the inconsistency of s_iso, which is described in this issue: https://github.com/oemof/oemof-thermal/issues/99

'stratified_thermal_storage.csv' contained 0.05 m as a value for s_iso. I've corrected it to 50 mm.

Issue #39, opened 2020-01-07
Module headers are not all the same format


jnnr:
While reading the api docs I noticed that the csp module header has a different form than the others. Maybe there are other inconsistencies? A quick check will be helpful.

jnnr:
Would be good to compare to oemof.solph's headers to ensure things are done right.

FranziPl:
I am not sure, what you mean. Is it another formatting (I cannot see that), or do you mean, I should write "concentrating_solar_power module" instead of "csp module".
If it is the second, I will change it.

FranziPl:
Shouldn't we have an API also for the flate plate collector?

jnnr:
> Shouldn't we have an API also for the flate plate collector?

That is true. I opened a separate issue #48 for that.

jnnr:
To clear things: With this issue I meant the missing headers in all modules with a short description and license info, like in oemof/oemof or tespy.

```
# -*- coding: utf-8

"""Module for loading a tespy network from saved state.

Use the method :func:`tespy.networks.network_reader.load_nwk` for importing
a network from a saved state.


This file is part of project TESPy (github.com/oemof/tespy). It's copyrighted
by the contributors recorded in the version control history of the file,
available from its original location tespy/networks/network_reader.py

SPDX-License-Identifier: MIT
"""
```

PR #49, opened 2020-01-14
Features/rename csp

FranziPl:
Change the header of the API for csp (will solve #39) and will additionally rename the example structure for the concentrating solor power example.


FranziPl:
@jnnr Could you take a quick look at this, so I can close this PR and maybe Issue 39, if this solves your problem in the Issue? Thanks.

jnnr (approved):

jnnr:
Looks good! It does solve the problem you found while reading #39, but not the whole issue.

PR #52, opened 2020-01-21
Add first tests for csp

FranziPl:
This PR will add tests for the concentrating solar power. It is a first sketch, so there are some tests missing, which I will add.


FranziPl:
I have a problem with the tests that are supposed to test the functions for a panda series. I managed to write a working assert statement. But it's not consistent with flake8. How can I change it according to flake8?

tests/test_concentrating_solar_power.py
```diff
@@ -0,0 +1,32 @@
...
+    assert res == 8.5381496824546241963970125
+
+
+def test_calculation_of_collector_irradiance_for_a_Series():
+    s = pd.Series([10, 20, 30], index=[1, 2, 3])
+    res = csp.calc_collector_irradiance(s, 0.9)
+    result = pd.Series(
+        [8.5381496824546242, 17.0762993649092484, 25.614449047363873],
+        index=[1, 2, 3])
+    assert res.eq(result).all() == True
```
jakob-wo:
```suggestion
    assert res.eq(result).all()
```

jakob-wo: This will make flake8 happy and it should be sufficient for the test because `res.eq(result).all()` will return either True or False itself.

tests/test_concentrating_solar_power.py
```diff
@@ -0,0 +1,32 @@
...
+    res = csp.calc_iam(-0.00159, 0.0000977, 50)
+
+    assert res == 0.8352499999999999
+
+
+def test_calculation_iam_for_a_Series():
+    s = pd.Series([10, 20, 30], index=[1, 2, 3])
+    res = csp.calc_iam(-0.00159, 0.0000977, s)
+    result = pd.Series([1.00613, 0.99272, 0.95977], index=[1, 2, 3])
+    assert res.eq(result).all() == True
```
jakob-wo:
```suggestion
    assert res.eq(result).all()
```

FranziPl:
Thanks @jakob-wo, now it works :)

FranziPl:
I added the tests in the PR #68, so I will close this PR

PR #54, opened 2020-01-22
Add solar thermal collector to apidocs

### LINKED ISSUE #48: Apidocs for solar thermal collector is missing
We should add the missing api documentation in api/oemof-thermal.rst
When doing that, we can adapt the order of the modules to alphabetical.

jnnr:
Aims to fix #48. Still open: Consistent naming, as addressed in the issue.


jnnr:
The build fails: https://readthedocs.org/api/v2/build/10305287%20%20.txt

Edit: Now it passes.

src/oemof/thermal/solar_thermal_collector.py
```diff
@@ -1,14 +1,17 @@
...
 
 """
 This module is designed to hold functions for calculating a solar thermal collector.
 
 This file is part of project oemof (github.com/oemof/oemof-thermal). It's copyrighted
-by the contributors recorded in the version control history of the file, available
-from its original location:
+by the contributors recorded in the version control history of the file,
+available from its original location:
 oemof-thermal/src/oemof/thermal/flat_plate_collector.py
```
FranziPl:
```suggestion
oemof-thermal/src/oemof/thermal/solar_thermal_collector.py
```

FranziPl (approved):
I renamed files and text according to #48. Should be ready to merge now.

PR #55, opened 2020-01-22
Unify module headers

### LINKED ISSUE #39: Module headers are not all the same format

jnnr:
Fixes #39.


jakob-wo (approved):
Looks good!

PR #56, opened 2020-01-22
Fixes/fix calc cops

jnnr:
The default mode=None leads to an error if the user does not set mode. Thus, this PR moves it to the required parameters.


jakob-wo:
Oh, I wasn't aware of that effect. Good to know and thanks for fixing it!

jakob-wo (approved):

PR #59, opened 2020-01-27
Correct file path

### LINKED ISSUE #58: Data in csp example cannot be found
The path in the scripts is given in uppercase letters, therefore the files cannot be found.

jnnr:
Fixes #58


FranziPl (approved):
Thanks for correct this.

Issue #13, opened 2019-09-30
We do not have a docs template for new components


jnnr:
The documentation of the new components should be consistent across the different components. To get there, we need some kind of template. Here a first draft:

* Scheme of the component  (e.g. flow diagram).
* Description.
* Link to docstrings for equations (as e.g. https://oemof.readthedocs.io/en/stable/oemof_solph.html#extractionturbinechp-component).
* Insightful output graphics (the same as in the example).

For the docstrings, align along the existing docstring template of oemof/oemof (see this issue https://github.com/oemof/oemof/issues/597)

Please extend this description with your points!

jakob-wo:
I propose to use a couple of subsections (sub-menu items). Have a look at the documentation of the module "Compressions Heat Pumps and Chillers" to get an impression.

In particular, I would like to draw your attention on the first subsection I would like to establish: "Motivation and possible application". Giving a reader the explanation why we actually developed this module and what we intended to use it for makes it much more ease to evaluate whether it is suitable or not for her/his own project.

jnnr:
> I propose to use a couple of subsections (sub-menu items). Have a look at the documentation of the module "Compressions Heat Pumps and Chillers" to get an impression.

I like the idea of subsections. They give structure and make it easy to get an overview. I would propose different names for the headlines, though. "Motivation and possible application" seems to refer to the motivation of the devs that lead them to develop that piece of code. I would expect such a thing in a paper, but not in the documentation of a module.

I guess the reader wants to know: What is the idea of these tools and how can I use them? The goal is to give a concise description of what the user can expect from the module. I suggest:

**Scope**
"This module was developed to provide ... for energy system optimizations with oemof.solph."
To which kind of systems do these tools apply to?

**Concept**
Scheme
Description of the concept
Equations

**Usage**
Overview table of parameters
Code snippets
Link(s) to example(s)
Insightful output graphics

**References**
Citation(s), if the concept is adapted from the literature.

jakob-wo:
> "Motivation and possible application" seems to refer to the motivation of the devs that lead them to develop that piece of code. 

In fact, that is my very intention. I want to tell people why we wrote that pieces of code. 

> I guess the reader wants to know: What is the idea of these tools and how can I use them? 

Definitely. I think telling people about our intention is a way to do so.

jnnr:
I guess we are aiming for the same: The docs should lower access barriers and be transparent. Of course it is essential to communicate the intention behind a tool.

[...]

PR #60, opened 2020-01-27
Unify headlines in components docs

jnnr:
Addresses #13.

@jakob-wo: Can you have a quick look and tell if you are fine?

### DIFF
--- a/docs/stratified_thermal_storage.rst
+++ b/docs/stratified_thermal_storage.rst
@@ -8,7 +8,7 @@ Scope
 _____
 
 This module was developed to implement a simplified model of a large-scale
-senible heat storage with ideal stratification for energy systems optimizations
+sensible heat storage with ideal stratification for energy system optimization
 with oemof.solph.
 
 
--- a/docs/stratified_thermal_storage.rst
+++ b/docs/stratified_thermal_storage.rst
@@ -4,9 +4,11 @@
 Stratified thermal storage
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Scope
+_____
 
 Concept
--------
+_______
 
 A simplified 2-zone-model of a stratified thermal energy storage.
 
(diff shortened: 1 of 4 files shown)


docs/stratified_thermal_storage.rst
```diff
@@ -7,6 +7,11 @@ Stratified thermal storage
 Scope
 _____
 
+This module was developed to implement a simplified model of a large-scale
+senible heat storage with ideal stratification for energy systems optimizations
```
jakob-wo:
```suggestion
sensible heat storage with ideal stratification for energy systems optimizations
```

jakob-wo: 1) A typo.
2) I would prefere: energy system(s) optimizations --> energy system optimizations.

docs/stratified_thermal_storage.rst
```diff
@@ -7,6 +7,11 @@ Stratified thermal storage
 Scope
 _____
 
+This module was developed to implement a simplified model of a large-scale
```
jakob-wo, line 10: is it explicitly made for _large-scale_ storages only? If it could be applied on smaller storages as well I would prefere not to mention the _large-scale_ here.

jakob-wo (changes requested):
Looks good! Just one typo to correct :)

docs/stratified_thermal_storage.rst
jnnr: Hi there! It can be applied to any size of storage, however the assumption of ideal stratification might not be valid for smaller ones.

docs/stratified_thermal_storage.rst
jnnr: Thanks, I corrected that :)

jakob-wo (approved):
ok! :)

Issue #65, opened 2020-02-05
Bugs / inconsistencies in the solar thermal collector documentation


MaGering:
I found bugs / inconsistencies in the solar thermal collector documentation.

1. Equations are missing due to a wrong path in the solar_thermal_collector.rst file
2. There is a name inconsistency between the symbol of collectors inlet temperature in the equation → Index: "collector in", the symbol convention in the documentation → Index: "coll" and the collectors inlet temperature in the table → Index: "Collector, in"

MaGering:
Solved by merging fix/docs-sections into dev

PR #66, opened 2020-02-05
Solve bugs / inconsistencies in solar thermal collector documentation

MaGering:
This PR contains three changes of the solar thermal collector documentation:

- Bugfix of missing equation due to wrong path

- symbol of collectors inlet temperature in equation

- symbol of collectors inlet temperature in table

#65


FranziPl (approved):
Thanks for finding and correcting these errors.

PR #68, opened 2020-02-12
Features/add csp method

FranziPl:
To validate the component for the concentrating solar power, we requested data from a real power plant. They characterise their collectors in a different way, so I added a second method to calculate the thermal losses and the efficiency. I changed the examples accordingly.

### DIFF
--- a/docs/concentrating_solar_power.rst
+++ b/docs/concentrating_solar_power.rst
@@ -83,7 +83,7 @@ These arguments are used in the formulas of the function:
     :math:`E^*_{coll}`        :py:obj:`irradiance_on_collector`                   Irradiance which hits collectors surface
                                                                                   before losses because of dirtiness are considered
 
-    :math:`X`                 :py:obj:`cleaniness`                                Cleanliness of the collector (between 0 and 1)
+    :math:`X`                 :py:obj:`cleanliness`                                Cleanliness of the collector (between 0 and 1)
 
     :math:`\kappa`            :py:obj:`iam`                                       Incidence angle modifier
 
--- a/docs/concentrating_solar_power.rst
+++ b/docs/concentrating_solar_power.rst
@@ -81,7 +81,7 @@ These arguments are used in the formulas of the function:
 
     :math:`E^*_{coll}`        :py:obj:`irradiance_on_collector`                   Irradiance which hits collectors surface
 
-    :math:`X`                 :py:obj:`x`                                         Cleanliness of the collector (between 0 and 1)
+    :math:`X`                 :py:obj:`cleaniness`                                Cleanliness of the collector (between 0 and 1)
 
     :math:`\kappa`            :py:obj:`iam`                                       Incidence angle modifier
 
--- a/docs/concentrating_solar_power.rst
+++ b/docs/concentrating_solar_power.rst
@@ -77,9 +77,11 @@ These arguments are used in the formulas of the function:
     ========================= =================================================== ===========
     symbol                    argument                                            explanation
     ========================= =================================================== ===========
-    :math:`E_{coll}`          :py:obj:`collector_irradiance`                      Irradiance on collector after all losses
+    :math:`E_{coll}`          :py:obj:`collector_irradiance`                      Irradiance on collector considering all losses
+                                                                                  including losses because of dirtiness
 
     :math:`E^*_{coll}`        :py:obj:`irradiance_on_collector`                   Irradiance which hits collectors surface
+                                                                                  before losses because of dirtiness are considered
 
     :math:`X`                 :py:obj:`cleaniness`                                Cleanliness of the collector (between 0 and 1)
 
--- a/src/oemof/thermal/concentrating_solar_power.py
+++ b/src/oemof/thermal/concentrating_solar_power.py
@@ -282,9 +282,9 @@ def calc_eta_c(eta_0, c_1, c_2, iam,
 
     """
     if loss_method == 'Janotte':
-        delta_t = (temp_collector_inlet + temp_collector_outlet) / 2 - temp_amb
-        eta_c = eta_0 * iam - c_1 * delta_t / collector_irradiance - c_2\
-            * delta_t ** 2 / collector_irradiance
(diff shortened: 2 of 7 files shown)


src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -265,16 +272,28 @@ def calc_eta_c(eta_0, c_1, c_2, iam,
...
-    delta_t = (temp_collector_inlet + temp_collector_outlet) / 2 - temp_amb
-    eta_c = pd.Series()
-    for index, value in collector_irradiance.items():
-        if value > 0:
-            eta = eta_0 * iam[index] - c_1 * delta_t[
-                index] / value - c_2 * delta_t[index] ** 2 / value
-            if eta > 0:
-                eta_c[index] = eta
+    if loss_method == 'Janotte':
+        delta_t = (temp_collector_inlet + temp_collector_outlet) / 2 - temp_amb
```
jnnr: delta_temp would be more consistant.

FranziPl: I changed it

FranziPl:
Because of developing two topics at the same time, I have more changes in this PR, than the header indicates:

- I added a second method
- I added tests for the csp
- I changed the part of the code for the calc_eta_c to a better kind.

@jnnr Could you please review this PR again, so I can close it? Thank you.

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -57,9 +60,7 @@ def csp_precalc(df, periods,
         degrees East of North
     x: numeric
         Cleanliness of the collector (between 0 and 1).
-    a_1: numeric
-        Parameter for the incident angle modifier.
-    a_2: numeric
+    a_1, a_2, a_3, a_4, a_5, a_6: numeric
```
jnnr, line 63: What are a_1, ..., a_6? Would be good to give some information. Looks quite obscure, this list of six numeric parameters.

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -113,6 +117,9 @@ def csp_precalc(df, periods,
     https://doi.org/10.21105/joss.00884
     """
 
+    if loss_method != "Janotte" and loss_method != "Andasol":
```
jnnr:
```suggestion
    if loss_method not in ['Janotte', 'Andasol']:
```

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -113,6 +117,9 @@ def csp_precalc(df, periods,
...
+        raise ValueError('loss_method should be "Janotte" or "Andasol"')
```
jnnr:
```suggestion
        raise ValueError("loss_method should be 'Janotte' or 'Andasol'")
```

jnnr:
Some more comments:

* in concentrating_solar_power.py:
  * The dataframe df already has a timeindex, which is internally replaced. This can lead to errors. 
  * Passing df and then telling which column means what is kind of complicated.
    I mean these 3:
    date_col: string, default: 'date'
    irradiance_col: string, default: 'E_dir_hor'
    temp_amb_col: string, default: 'temp_amb'

Suggestion: Pass irradiance (E_dir_hor or dni, depending on irradiance_method) and temp_amb as pd.Series and check that it has a valid timeindex (which is necessary for the pvlib). E_dir_hor/dni should be passed as kwargs, and depending on the method it should be checked that the right keyword argument is passed.

``` python
def csp_precalc(
                lat, long, timezone,
                collector_tilt, collector_azimuth, cleanliness,
                eta_0, c_1, c_2,
                temp_collector_inlet, temp_collector_outlet,
                temp_amb,
                a_1, a_2, a_3=0, a_4=0, a_5=0, a_6=0,
                loss_method='Janotte',
                irradiance_method='horizontal',
                **kwargs):
                """ ... """
               required_dict = {'horizontal': 'E_hor_ir', 'other': 'dni'}

               irradiance_required = required_dict[irradiance_method]

               if not irradiance_required in kwargs:
                   raise AttributeError(f"'{irradiance_required}' necessary for {irradiance_method} is not provided")

               irradiance = kwargs.get(irradiance_required)

```

* parameter names
  * tz: why not call it timezone?
  * x: why not call it cleanliness? I would be ok with x however if it is an established convention to name it like this.

* I do not quite understand the difference btw. collector_irradiance and irradiance_on_collector. The explanation in the table does not make it clear to me. Also, we should check how the pictures in the docs help understanding these (#76)

jnnr:
There is a typo that should lead to a failure in line 163:

pvlib.irradiance.beam_coponent

jnnr (approved):

src/oemof/thermal/concentrating_solar_power.py
FranziPl: I added an explanation, in which case they are needed. Apart from that i consider "Parameter for the incident angle modifier" to be sufficient

docs/concentrating_solar_power.rst
```diff
@@ -77,11 +77,13 @@ These arguments are used in the formulas of the function:
...
     ========================= =================================================== ===========
-    :math:`E_{coll}`          :py:obj:`collector_irradiance`                      Irradiance on collector after all losses
+    :math:`E_{coll}`          :py:obj:`collector_irradiance`                      Irradiance on collector considering all losses
+                                                                                  including losses because of dirtiness
 
     :math:`E^*_{coll}`        :py:obj:`irradiance_on_collector`                   Irradiance which hits collectors surface
+                                                                                  before losses because of dirtiness are considered
 
-    :math:`X`                 :py:obj:`x`                                         Cleanliness of the collector (between 0 and 1)
+    :math:`X`                 :py:obj:`cleaniness`                                Cleanliness of the collector (between 0 and 1)
```
FranziPl:
```suggestion
    :math:`X`                 :py:obj:`cleanliness`                                Cleanliness of the collector (between 0 and 1)
```

FranziPl:
Thanks a lot for your input! It was very helpful.

PR #70, opened 2020-02-16
Correct default values in stratified_thermal_storage.py

MaGering:
With this pull request the default values for the density and capacity are corrected in the functions 

- calculate_capacities() and 
- calculate_losses() 

in stratified_thermal_storage.py. As a simplification the values are calculated with CoolProp for a constant temperature of 80 °C.

### DIFF
--- a/src/oemof/thermal/stratified_thermal_storage.py
+++ b/src/oemof/thermal/stratified_thermal_storage.py
@@ -112,11 +112,13 @@ def calculate_capacities(
 
     heat_capacity: numeric
         Average specific heat capacity of storage medium [J/(kg*K)]
-        Default values calculated with CoolProp for a temperature of 80 °C as a simplifying assumption
+        Default values calculated with CoolProp for a temperature of 80 °C
+        as a simplifying assumption
 
     density : numeric
         Average density of storage medium [kg/m3]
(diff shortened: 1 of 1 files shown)


MaGering:
This approach has already been discussed in the following conversation:

https://github.com/oemof/oemof-thermal/pull/19#discussion_r352484326

src/oemof/thermal/stratified_thermal_storage.py
```diff
@@ -112,9 +112,11 @@ def calculate_capacities(
 
     heat_capacity: numeric
         Average specific heat capacity of storage medium [J/(kg*K)]
+        Calculated with CoolProp at a constant temperature of 80 °C as a simplification
```
jnnr:
```suggestion
        Default values calculated with CoolProp for a temperature of 80 °C as a simplifying assumption
```

jnnr (commented):
Looks good, thanks! I suggest to change the text slightly, referring to the default values.

src/oemof/thermal/stratified_thermal_storage.py
MaGering: Thank you for your suggestion! I've adjusted the text accordingly.

jnnr (approved):

PR #71, opened 2020-02-16
Add nominal_storage_capacity in max_storage_level and min_storage_level

MaGering:
This pull request includes the multiplication with the nominal_storage_capacity in the calculation of the return values "max_storage_level" and "min_storage_level" (function calculate_capacities()) accordingly to the procedure in the docs:

https://oemof-thermal.readthedocs.io/en/latest/stratified_thermal_storage.html

### DIFF
--- a/src/oemof/thermal/stratified_thermal_storage.py
+++ b/src/oemof/thermal/stratified_thermal_storage.py
@@ -131,8 +131,8 @@ def calculate_capacities(
     nominal_storage_capacity = volume * heat_capacity * density * (temp_h - temp_c)
     nominal_storage_capacity *= 1 / 3600  # J to Wh
     nominal_storage_capacity *= 1e-6  # Wh to MWh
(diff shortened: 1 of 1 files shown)


jnnr:
'minimum_storage_level' and 'maximum_storage_level' are defined relative, they are between 0 and 1 (as explained in the docstring of GenericStorage https://oemof.readthedocs.io/en/stable/api/oemof.solph.html#oemof.solph.components.GenericStorage). There is no need to multiply the 'nominal_storage_capacity'.

MaGering:
I see! Thank you for explaining. I also completely overlooked "relative to nominal storage capacity" in calculate_capacities(...) docstring.

Would it in this case be helpful to clarify that in the documentation?

jnnr:
> Would it in this case be helpful to clarify that in the documentation?

If there is a way to do so, sure. Where would you suggest to clarify it? If you have a suggestion, please open a new issue and close this PR.

Issue #50, opened 2020-01-14
The module name "chp" does not comply with the (unwritten) naming standard


jakob-wo:
First, the module name `chp.py` is an abrevation unlike the other names where "thermal_storage" is used instead of TES, and "concentrating_solar_power" instead of CSP. In my opinion it would be nice to rename the module `combined_power_and_heat.py`.

Second, we could even use the renaming to find a more appropiate and less missleading name for the module. "CHP" indicates that the module can be used to model the component CHP power plant because such power plant are often called CHP itself. I suggest we use a name that does not refer to a actual machine/unit - such as `cogeneration.py`.

Do you agree to renaming the modue `combined_power_and_heat.py`?
What do you think about changing the name to `cogeneration.py`?

FranziPl:
What about emission_allocation_of_chp? too long?

jnnr:
I like `cogeneration.py`, as it is shorter and is more general. Maybe `cogeneration_emission_allocation.py` would be fine, too? To decide, I would suggest to think about what developments are planned in the near future that could fit in this module, and choose a suitable name.

FranziPl:
I'm fine with `cogeneration.py`, but I would prefere `cogeneration_emission_allocation.py` as the former implies, that it will calculate the generation and not the emission allocation.

jnnr:
I thought about it again and it seems to me that cogeneration_emission_allocation is a bit to specific. It could describe the function in the module itself and does not give room for more. Hence I would vote for cogeneration.py.

PR #72, opened 2020-02-17
Features/cogeneration

jnnr:
This PR renames the existing chp module to cogeneration, according to #50. I also added an example that creates a similar plot to the one in the publication by Mauch. Optionally, we could also add an example that shows how to use emission_allocation together with an oemof model. What do you think, @oemof/oemof-heat ?

TODO: 
* [x] Add an example showing how the function works.
* [x] Write some sentences in the docs.
* [x] Test if function works with series as input as well.


jakob-wo:
Great! The plot looks good and we should use it in the documentation. This is also another ToDo I would add:

- [ ] Add documentation for module _cogeneration_

I like the idea of an example that illustrates the usage with solph. I will do that and suggest an example in the next days.

- [ ] Add example with solph (can be done independently from this PR)

One last thing: I would rename the example that creates the plot. Right now it is called `cogeneration.py` - just as the module itself. We could use a more specific name. What about the name `emission_allocation_methods.py`?

- [x] Decide whether to rename first example

jnnr:
> Great! The plot looks good and we should use it in the documentation. This is also another ToDo I would add:
> 
>     * [ ]  Add documentation for module _cogeneration_

I set up a documentation page for cogeneration. There is still some text that has to be written.

> One last thing: I would rename the example that creates the plot. Right now it is called `cogeneration.py` - just as the module itself. We could use a more specific name. What about the name `emission_allocation_methods.py`?
> 
>     * [ ]  Decide whether to rename first example

I like it. Adopted your suggestion.

I also sorted the docs and apidocs of the modules alphabetically and structured the navigation bar.

jakob-wo (approved):

jakob-wo:
You could also add the PEE calculation (finnish method) to the docs.

Apart from that, it looks good and can be merged.

PR #75, opened 2020-02-20
Add function tests for solar thermal collector

### LINKED ISSUE #34: There are no tests

MaGering:
Adds first sketch of tests for the module solar thermal collector. 
Related issue: https://github.com/oemof/oemof-thermal/issues/34


c-moeller:
@MaGering Please add the external calculated values to the tests.
@FranziPl Please review the code.
Thanks a lot!

MaGering:
> @MaGering Please add the external calculated values to the tests.

I've calculated the values in an external script. They are equal to the values calculated within solar_thermal_collector.py, which I used in test_functions.py as placeholders so far. This means that test_functions.py does not change.

FranziPl (approved):
Looks fine to me.

jnnr:
This one seems ready to merge. @MaGering: Can you please resolve the merge conflict?

PR #76, opened 2020-02-21
Features/improve docu pics

### LINKED ISSUE #61: Schematic of components does not follow the same logic for all components
The schematics of the 4 components in this package follow a different logic. Where the ones for stratified_thermal_storage and compression_heat_pumps_chillers show the energy/mass flows and balance in a diagram, the solar_thermal_collector and concentrating_solar_power rather show information flows.

If I remember correctly, we said that we would like to follow the former concept (energy/mass flows). In my opinion, this is useful to know what we are talking about. It makes it easier to understand the concept of the component that is to be represented. 

If you agree, this would mean that the collector and csp module would get new schematics in the next release.

Maybe, the information flow picture is helpful in its own right. It could then move further down in the text. Also, keep in mind that this might change when facades are implemented.

Also, the schematics of 

* [ ] concentrating solar power (@FranziPl),
* [ ] solar collector (@c-moeller) and 
* [x] stratified thermal storage  (@jnnr) resolved in #42 

do not have a caption. Please add a caption similar to the one in https://oemof-thermal.readthedocs.io/en/latest/compression_heat_pumps_and_chillers.html

FranziPl:
This PR will add a schematic of the csp, which will match the schematics of the other components.
It will solve #61.
@jakob-wo or @jnnr: Can either of you look at it?


jnnr:
Thanks, that looks good! Really helpful to have that image.

Some remarks:

* The image canvas is quite large, leading to a lot of blank space in the docs. You could shrink it to the size of the actual picture.
* Illustrating the geometrical losses with curly arrows suggests that the losses are somehow scattered/dissipated, which could be misleading.
* Q_coll is the difference between the heat entering and the heat leaving the collector. Maybe rather write mass flows and temperatures to the arrows?
* The red arrow should be parallel to the contour of the mirror.

FranziPl:
I changed point 1, 3, and 4. I disagree to the second point, because I think that there are many arrows and information in this picture. And if I don't have a consistent form for the different types of loss, it would be confusing. Besides, I have followed your picture in this point.

FranziPl:
Point 2 is now also considered.
Ready to merge in my opinion. 
I would like to have another review.

c-moeller:
@FranziPl Do you plan such kind of pic for the solar thermal collector as well? Otherwise @MaGering could do this.

docs/concentrating_solar_power.rst
```diff
@@ -29,9 +29,24 @@ _________
...
+    :width: 60 %
+    :alt: concentrating_solar_power.svg
+    :align: center
+    :figclass: align-center
+
+    Fig.1: The energy flows at the collector.
+
+The geometrical losses, the optical losses at the collector, and the thermal
+losses in the collector are taking into account to calculate the heat, which
+the collector can provide.
```
c-moeller: What do you think about changing this text to something like the following (I took this from the specification sheet of the component):

> The direct normal radiation (E_dir) is reduced by geometrical losses (Q_loss,geom) so that only the collector radiation (E_coll*) hits the collector. Before the thermal power is absorbed by the absorber tube, also optical losses (Q_loss,opt), which can be reflection losses at the mirror, transmission losses at the cladding tube and absorption losses at the absorber tube, occur. The absorber finally loses a part of the absorbed heat output through thermal losses (Q_loss,therm).

It might be a bit easier to capture the picture.

FranziPl:
> 
> 
> @FranziPl Do you plan such kind of pic for the solar thermal collector as well? Otherwise @MaGering could do this.

I didn't plan it, but it is a svg file, to it is possible to adapt it.

docs/concentrating_solar_power.rst
jnnr: I agree. It would be helpful to explain the different processes that the incoming radiation undergoes until it is finally converted to usable heat a bit more, so the reader can understand the picture and relate that to the math following below.

jnnr:
Hey @FranziPl, thanks for the picture. It looks really good! Really helps understanding.

Just some small things: You could consider extending the text in the caption. In my experience, the pictures are the parts in a long text that you grasp first, so it is always good to ship it with some text that just explains what we see on the picture.

I would also agree with @c-moeller with their comment above.

I found some typos and I took the freedom to fix them. 

Do not forget to add a line to whatsnew (https://github.com/oemof/oemof-thermal/blob/dev/docs/whats_new/v0-0-2.rst) before you merge.

MaGering:
> I would like to have another review.

Unfortunately svg figures won't render in pdfLaTeX: https://github.com/oemof/oemof-thermal/pull/91

Hence a .png-file  of 'concentrating_solar_power.svg' needs to be embedded in 'concentrating_solar_power.rst'.

jnnr:
Should the losses in the image, math:`Q`, not rather be declared as  math:`\dot{Q}`?

MaGering:
I've added a svg schematic of the solar thermal collector to the folder _pics.

docs/concentrating_solar_power.rst
FranziPl: I integrated the text

FranziPl:
I am done. I changed the picture from svg to png, I added a line to the whats new file and I integrated the comments of jnnr and c-moeller.
If I didn't miss anything, it can be merged.

jnnr:
> Should the losses in the image, math:`Q`, not rather be declared as math:`\dot{Q}`?

I saw that this is adapted for the csp. I guess this should be adapted for the solar_thermal_collector as well, shouldn't it, @MaGering?

jnnr (changes requested):
As written in a previous comment, I think that in the image showing the solar thermal collector, Q has to become dot Q. Do you agree, or am I wrong, @MaGering?

FranziPl:
@MaGering did the requested change in #84 instead of this one, but it is done.

jnnr (approved):

jnnr:
> @MaGering did the requested change in #84 instead of this one, but it is done.

Ok. In this case I approve. From my view this seems ready to be merged.

FranziPl:
Travis doesn't work here, it is still waiting for a status.
I checked pytest and flake8 locally, so the branch is ok and could be merged.
But I have no idea, why travis has a problem here. Does anybody have an idea? Should I merge anyway?

PR #77, opened 2020-02-25
Add csp facade

FranziPl:
This PR should build a facade for the component csp.
So far, it is just a file, which shows, how an example with the facade should look. It doesn't work yet.

TODOs
* [x] pass a_1, a_2 instead of dict
* [x] increase test coverage by writing a test
* [x] resolve merge conflicts


examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+size_collector = 1000
+
+# busses
+bth = solph.Bus(label='thermal')
+bel = solph.Bus(label='electricity')
+
+#sources and sinks
+collector = facades.Collector(
+    label='solar_collector',
+    output=bth,
```
FranziPl:
```suggestion
    heat_bus=bth,
```

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+    electrical_input=bel,
```
FranziPl:
```suggestion
    electrical_bus=bel,
```

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+    electrical_consumption=elec_consumption,
```
FranziPl, line 69: This is the ratio of electricity (for the pump) per heat output. Maybe find a name that is clearer?

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+    peripheral_losses=eta_losses,
+    size=size_collector,
```
FranziPl: Rethink the name. Maybe aperture_area or something like this?

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+    loss_method='Janotte',
+    irradiance_method='horizontal',
+    latitude=23.614328,
+    longitude=58.545284,
+    timezone='Asia/Muscat',
+    collector_tilt=10,
+    collector_azimuth=180,
+    x=0.9,
```
FranziPl: cleanliness

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+    a_1=-0.00159,
+    a_2=0.0000977,
```
FranziPl: and a_3 ... a_6 in kwargs

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,125 @@
...
+    eta_0=0.816,
+    c_1=0.0622,
+    c_2=0.00023,
+    temp_collector_inlet=435,
+    temp_collector_outlet=500,
+    irradiance=input_data['E_dir_hor'],
```
FranziPl: E_dir_hor / dni in kwargs, depending on method.

FranziPl:
I have a first try for the facade here.
I copied the comparison of the results of the examples with and without facade from jnnr.

The facade is not working. The problem I have: I tried to use the facade reservoir from oemof-tabular, because it has a source, which is combined with another component.
Reservoir is a source with a storage. There is no bus for the source, but it is connected directly to the storage.
In my case, I have a transformer, which I want to combine with a source.
The reservoir works with only one bus. The transformer works with 2 busses. I tried to combine the facade for a transformer with the one for the reservoir, but I was not able to make the build_solph_components work.

As I changed the structur of the csp in the PR #84, i merged the new structur of the branch features/improve_code_csp into this branch. Therefore the PR #84 must first be merged so that this branch can be compared to the correct code.

jnnr:
I can have a review on Friday, if you want.

FranziPl:
Yes, please do it.

FranziPl:
The facade is working now. Maybe the style is a little rough, but it is working.
I created an example for the facade and adapted the existing plant example. They produce the same results. I didn't create an automatic comparison, but compared the produced csv-tables by hand. The comparison of the lp-files is not possible, as the facade can save on a bus.

FranziPl:
So far it just works for the method "Janotte". The second method still must be integrated (as there are different arguments, I haven't done it yet).

FranziPl:
I added the second method for the irradiance and the second method for the losses.
I didn't test the second method for the irradiance yet, but as it just call the csp_precalc_function, there should be no false calculation in the facade itself. And I checked, if it is callable.
I did check, id the second loss method is useable, but I didn't check the results yet as well. I might write an example for that, but I will not make it today.
So the PR is ready for review.

docs/whats_new/v0-0-2.rst
```diff
@@ -9,7 +9,8 @@ API changes
 New features
 ------------
 
-* 
+* A Facade for the concentrating solar power is now available. There is an
+example, which shows, how the facade is used.
```
jnnr:
```suggestion
example which shows how the facade is used.
```

examples/concentrating_solar_power/csp_collector_plot_example.py
```diff
@@ -22,7 +22,14 @@
 
 dataframe = pd.read_csv('csp_data/data_Muscat_22_8.csv')
 dataframe['Datum'] = pd.to_datetime(dataframe['Datum'])
+dataframe.set_index('Datum', inplace=True)
+dataframe.index = dataframe.index.tz_localize(tz='Asia/Muscat')
```
jnnr: What happens if the user forgets to do this step? I guess that the time zone is necessary for pvlib. Does the respective function raise an error if this is missing or does it fail without the user noticing it?

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,158 @@
...
+costs_electricity = 1000
+conversion_factor_turbine = 0.4
+size_collector = 1000
+
+# busses
+bth = solph.Bus(label='thermal')
+bel = solph.Bus(label='electricity')
+
+#sources and sinks
+collector = facades.Collector(
```
jnnr: Is the separation between collector and turbine intended? As I remember previous discussions, the source and transformer were meant to be put into one facade class. Looking at the [scheme](https://github.com/oemof/oemof-thermal/blob/features/facade_csp/docs/_pics/scheme.png), this seems to have been the plan.

Here, they are separated. Is this as you decided it, or do you still think it would be nice to include the turbine in the facade, but there are technical obstacles to do so?

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,158 @@
...
+    additional_losses=additional_losses,
+    aperture_area=size_collector,
+    loss_method='Janotte',
+    irradiance_method='horizontal',
+    latitude=latitude,
+    longitude=longitude,
+    collector_tilt=collector_tilt,
+    collector_azimuth=collector_azimuth,
+    cleanliness=cleanliness,
+    aas={
```
jnnr: Is there an advantage in passing these as a dict? Seems a bit funny to have a separate dictionary here. You could do the same as in csp_precalc, defining default values for a_3 ... a_6 and have one argument for every parameter.

jnnr:
Running the example_csp_facade, I get an error

``` bash
  File ".virtualenvs/oemof/lib/python3.6/site-packages/oemof/solph/models.py", line 73, in __init__
    self.es.timeindex.freq.nanos / 3.6e12)
AttributeError: 'NoneType' object has no attribute 'nanos'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "example_csp_facade.py", line 131, in <module>
    model = solph.Model(energysystem)
  File ".virtualenvs/oemof/lib/python3.6/site-packages/oemof/solph/models.py", line 267, in __init__
    super().__init__(energysystem, **kwargs)
  File ".virtualenvs/oemof/lib/python3.6/site-packages/oemof/solph/models.py", line 79, in __init__
    raise AttributeError(msg)
AttributeError: No valid time increment found. Please pass a valid timeincremet parameter or pass an EnergySystem with a valid time index. Please note that a valid timeindex need to have a 'freq' attribute.
```

I can fix it by adding

```date_time_index.freq = 'H'```

to the beginning of the script. Did you get that error as well? Maybe it depends on the oemof version.

**Comment by FranziPl** [comment]
I don't have this error. Maybe it is the pandas version?
I have pandas 0.23.4 and oemof 0.3.1.

**Comment by jnnr** [comment]
Did you compare the lp-files of the facade example with the one without facade? It is a good way to check that you get the same results for both approaches. I compared, but got different lp-files. This is because the two examples are set up in a slightly different way. Would be a good check to try to reproduce the previous runs with the new facade.

**Comment by jnnr** [comment]
> I don't have this error. Maybe it is the pandas version?
> I have pandas 0.23.4 and oemof 0.3.1.

Yes, pandas it is! I had 0.24.2. When I switch to 0.23.4, everything works without setting the freq attribute.

File: examples/concentrating_solar_power/example_csp_facade.py (hunk 10 above)
Comment (jnnr) [reply examples/concentrating_solar_power/example_csp_facade.py]: Maybe I got it wrong. As I learned when looking into the facade, it wraps a Transformer and a Source. So the collector includes the collector and its pump, but not the turbine. I guess this was always the plan.

Do you plan to update and include the scheme to show it in the docs? Could be helpful to understand the workings of the facade. However, this can be a separate step in a new PR.

**Comment by jnnr** [comment]
Test coverage decreases a lot with this PR. We should add more tests before merging. In test_constraints, such a test can be added. If you need help, tell me!

**Comment by FranziPl** [comment]
I did compare the lp files, but removed this comparison later. It is not possible, because the facade uses one less bus (this is one of the advantages of the facade, it makes the model less complicated.

File: examples/concentrating_solar_power/example_csp_facade.py (hunk 10 above)
Comment (FranziPl) [reply examples/concentrating_solar_power/example_csp_facade.py]: Like you said in the second comment, it was not the plan to include the turbine. I agree, a scheme would help, I am not sure, when I will find the time.

File: examples/concentrating_solar_power/example_csp_facade.py (hunk 11 above)
Comment (FranziPl) [reply examples/concentrating_solar_power/example_csp_facade.py]: To be honest, I just didn't find another way to deal with a different number of arguments in the class, which are given to the precalc-function. I don't know, how to ask for a_3 in the kwargs without raising an error, if it is not given by the user. Could you help me with that, please?

File: examples/concentrating_solar_power/csp_collector_plot_example.py (hunk 9 above)
Comment (FranziPl) [reply examples/concentrating_solar_power/csp_collector_plot_example.py]: The user has to do the first step. But this is required in the API, which says "time indexed series".
The second one depends on the users data. My data are given with local time, so I have to change/correct them. If the user already has his data with the UTC time or an dataframe, which has a correct time stamp (including timezone).

**Comment by FranziPl** [comment]
> 
> 
> > I don't have this error. Maybe it is the pandas version?
> > I have pandas 0.23.4 and oemof 0.3.1.
> 
> Yes, pandas it is! I had 0.24.2. When I switch to 0.23.4, everything works without setting the freq attribute.

I didn't find a possibility to set the frequence in the same step, so the additional step you made is still required for the newer pandas versions. Sould I jsut add this line into my example?

**Comment by FranziPl** [comment]
> 
> 
> Test coverage decreases a lot with this PR. We should add more tests before merging. In test_constraints, such a test can be added. If you need help, tell me!

Yes, I do, I will write you a mail.

**Comment by FranziPl** [comment]
I thought, it is ready to merge now, but Travis fails here, while it works locally.
Do you know why? @jnnr : Is it the same reason, why you skip your tests in the moment (because I use the same functionality to test the facade.)

**Comment by jnnr** [comment]
> I thought, it is ready to merge now, but Travis fails here, while it works locally.
> Do you know why? @jnnr : Is it the same reason, why you skip your tests in the moment (because I use the same functionality to test the facade.)

The tests of StratifiedThermalStorage are skipped as they rely on a feature in oemof.solph that is not yet released, so that should not apply to the Collector. For me, it fails locally as well.

The reason it fails is a difference in the last decimal values of a float.

= -75650.812223268411 vs. = -75650.81222326828

Not a big difference, but still for some reason.

**Comment by jnnr** [comment]
You might want to rename the facade. `Collector` is general and could be confused with the flat plat collector. Also, we might consider renaming `SolarThermalCollector` to `FlatPlateCollector` as it is more specific. What do you think, @FranziPl, @c-moeller?

File: tests/test_constraints.py [hunk 12]
```diff
@@ -0,0 +1,221 @@
...
+
+    def test_csp_collector_facade(self):
+        """Constraint test of a csp collector.
+        """
+        bus_heat = solph.Bus(label='bus_heat')
+        bus_el = solph.Bus(label='bus_el')
+
+        d = {
+            'Datum': [
+                '01.02.2003 09:00', '01.02.2003 10:00', '01.02.2003 11:00',
```
jnnr: Remember that the number of timesteps, i.e. periods, is 3 for the EnergySystem that is set up. You do not need that much data here.

examples/concentrating_solar_power/csp_plant_example.py
```diff
@@ -13,30 +13,24 @@
...
 import oemof.outputlib as outputlib
 import matplotlib.pyplot as plt
 
-
 # set results path
 base_path = os.path.dirname(os.path.abspath(os.path.join(__file__)))
 
-results_path = os.path.join(base_path, 'results')
+results_path = os.path.join(base_path, 'results/')
+lp_path = os.path.join(base_path, 'lp_files/')
```
c-moeller, line 24:
```suggestion
```

c-moeller: Do you still need the lp_path?

examples/concentrating_solar_power/csp_plant_example.py
```diff
@@ -134,34 +135,33 @@
...
-
 model.solve(solver='cbc', solve_kwargs={'tee': True})
 
-# filename = (path + '/lp_files/'
-#             + 'CSP_Test.lp')
-# model.write(filename, io_options={'symbolic_solver_labels': True})
+# if not os.path.exists(lp_path):
+#         os.mkdir(lp_path)
+# model.write((lp_path + 'csp_model.lp'),
+#             io_options={'symbolic_solver_labels': True})
```
c-moeller, line 151: Remove?

examples/concentrating_solar_power/csp_plant_example.py
```diff
@@ -13,30 +13,24 @@
 from oemof import solph
 from oemof.tools import economics
 from oemof.thermal.concentrating_solar_power import csp_precalc
+
 import oemof.outputlib as outputlib
 import matplotlib.pyplot as plt
 
-
 # set results path
```
c-moeller:
```suggestion
# set path
```

examples/concentrating_solar_power/example_csp_facade.py
jnnr: Ok.

examples/concentrating_solar_power/example_csp_facade.py
jnnr: I agree, that would be helpful!

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,157 @@
...
+costs_storage = economics.annuity(20, 20, 0.06)
+costs_electricity = 1000
+conversion_factor_turbine = 0.4
+size_collector = 1000
+
+# busses
+bth = solph.Bus(label='thermal')
+bel = solph.Bus(label='electricity')
+
+#sources and sinks
```
c-moeller:
```suggestion
# collector
```

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,157 @@
...
+    a_1=a_1,
+    a_2=a_2,
+    eta_0=eta_0,
+    c_1=c_1,
+    c_2=c_2,
+    temp_collector_inlet=temp_collector_inlet,
+    temp_collector_outlet=temp_collector_outlet,
+    temp_amb=input_data['t_amb'],
+    irradiance=input_data['E_dir_hor'])
+
```
c-moeller, line 88:
```suggestion

# sources and sinks
```

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,157 @@
...
+        fixed=True,
+        actual_value=input_data['ES_load_actual_entsoe_power_statistics'],
+        nominal_value=1)})
+
+excess = solph.Sink(
+    label='excess',
+    inputs={bth: solph.Flow()})
+
+# transformer and storages
+
```
c-moeller:
```suggestion
```

examples/concentrating_solar_power/example_csp_facade.py
```diff
@@ -0,0 +1,157 @@
...
+storage = solph.components.GenericStorage(
+    label='storage_th',
+    inputs={bth: solph.Flow()},
+    outputs={bth: solph.Flow()},
+    loss_rate=cap_loss,
+    inflow_conversion_factor=conversion_storage,
+    outflow_conversion_factor=conversion_storage,
+    investment=solph.Investment(ep_costs=costs_storage))
+
```
c-moeller, line 124:
```suggestion

 # build the system and solve the problem  
```

examples/concentrating_solar_power/csp_plant_example.py
FranziPl: If somebody wants to have a lokk at the lp_file (or also me) for test reasons, it would be needed.

examples/concentrating_solar_power/csp_plant_example.py
FranziPl: The same as in the answer before: Could be useful for understanding and testing. Not necessary.

FranziPl:
> 
> 
> You might want to rename the facade. `Collector` is general and could be confused with the flat plat collector. Also, we might consider renaming `SolarThermalCollector` to `FlatPlateCollector` as it is more specific. What do you think, @FranziPl, @c-moeller?

I am not sure about the naming. I agreee in general, we need different names. But the flat plate collector is also a solar thermal collector, so maybe it would be confusing, too.

jnnr:
> > You might want to rename the facade. `Collector` is general and could be confused with the flat plat collector. Also, we might consider renaming `SolarThermalCollector` to `FlatPlateCollector` as it is more specific. What do you think, @FranziPl, @c-moeller?
> 
> I am not sure about the naming. I agreee in general, we need different names. But the flat plate collector is also a solar thermal collector, so maybe it would be confusing, too.

Revisiting this, it seems consistent to stick with the name `SolarThermalCollector`. For CSP though, is there a good name that is more specific than just `Collector`?

FranziPl:
> 
> 
> > > You might want to rename the facade. `Collector` is general and could be confused with the flat plat collector. Also, we might consider renaming `SolarThermalCollector` to `FlatPlateCollector` as it is more specific. What do you think, @FranziPl, @c-moeller?
> > 
> > 
> > I am not sure about the naming. I agreee in general, we need different names. But the flat plate collector is also a solar thermal collector, so maybe it would be confusing, too.
> 
> Revisiting this, it seems consistent to stick with the name `SolarThermalCollector`. For CSP though, is there a good name that is more specific than just `Collector`?

parabolic trough collector would be possible. I will name it like this.

examples/concentrating_solar_power/example_csp_facade.py
FranziPl: I thought about it and I think, it describes well, what is ment.

FranziPl:
@jnnr:
I updated oemof and pandas and the tests still pass locally at my machine. 
So I have no idea, what I could do, to make the tests work at Travis, too.

examples/concentrating_solar_power/csp_plant_example.py
```diff
@@ -48,6 +42,13 @@
 temp_collector_inlet = 435
 temp_collector_outlet = 500
 
+# input data
+dataframe = pd.read_csv(data_path + 'data_csp_plant.csv').head(periods)
+dataframe['Datum'] = pd.to_datetime(dataframe['Datum'])
+dataframe.set_index('Datum', inplace=True)
+dataframe.index = dataframe.index.tz_localize(tz='Asia/Muscat')
```
MaGering:
```suggestion
dataframe.set_index('Datum', inplace=True)
dataframe = dataframe.asfreq('H')
dataframe.index = dataframe.index.tz_localize(tz='Asia/Muscat')
```

I somehow also get the same error mentioned by @jnnr in the comment https://github.com/oemof/oemof-thermal/pull/77#issuecomment-608494618 running the script if I don't assign 'H' to freq.

I have oemof 0.3.2 and pandas 0.25.3 versions.

FranziPl: Thanks, I just changed this in one of the two example files.

jnnr:
I fixed the test locally at my machine and pushed the changes. The change is only in the last decimal digits of one number, which I adapted in the lp file. Now the CI test passes. Can you cross check this?

FranziPl:
> 
> I fixed the test locally at my machine and pushed the changes. The change is only in the last decimal digits of one number, which I adapted in the lp file. Now the CI test passes. Can you cross check this?

Now it fails locally at my machine, if I use pytest.

c-moeller:
The test also passes at my machine. I'm going to merge.

Issue #46, opened 2020-01-13
Docs pdf has bugs


jnnr:
On oemof-thermal.readthedocs.io, you can download document versions of the docs if you click on latest:

![Screenshot_2020-02-25_15-06-28](https://user-images.githubusercontent.com/32454596/75254325-80001000-57e0-11ea-8d39-4c1c32feebae.png)

There is a pdf, an epub and a html version.

There are the following bugs:

* The title is 'oemof heat documentation', not oemof.thermal. Should be fixed by changing conf.py
* .svg-files do not seem to render.

In .epub version: 
* Image size leads to pixelized appearance
* .svg images are not properly handled either

jnnr:
This could help: https://github.com/godotengine/godot-docs/issues/1352

jnnr:
In .epub version: 
* Image size leads to pixelized appearance
* .svg images are not properly handled either

jnnr:
@MaGering: Can you have a look at this one? Not urgent, but should be fixed before the next release in mid of March.

PR #81, opened 2020-03-01
Change "oemof heat" to "oemof.thermal" in docs conf.py

MaGering:
This PR changes the target name and title of documents that are output within conf.py.

It has been changed for

- LaTeX
- manual page
- Texinfo
- Epub

Related issue: https://github.com/oemof/oemof-thermal/issues/46


jnnr:
Thanks for taking this up. I told readthedocs to create a build for this branch. Have a look:

https://oemof-thermal.readthedocs.io/en/fix-docs-sections/

It seems that the pdf, epub and html are not created any more. Can you try to fix that?

MaGering:
I've tried to use extensions (sphinxcontrib.inkscapeconverter and sphinx.ext.imgconverter) for .svg-files, but the images could not render in pdflatex. I got errors using the extensions.
So i changed the .svg-files to .pdf. This however led to .pdf-files not rendering in epub.
That's why i changed the files to .png (paying attention to the resolution). Now they seem to render everywhere.

jnnr:
Hey, thanks for trying all these!

If readthedocs / sphinx cannot be configured to handle svg for the pdf and epub documents, converting the images to png seem to be the solution to take. Still, it would be good to keep the svg versions in the same directory in case someone wants to adapt a figure.

We could also add a little readme that explains the problem, the things you tried and how we handle this now: If your original pictures are svg, commit them together with a png version that is actually included in the docs. Maybe at a later point in time we can revisit this and there will be a better solution. But for now, it is ok if it fixes the bugs.

MaGering:
In order to clean this PR the issues in the docs have been separated in:

1. naming issue:
handled in PR https://github.com/oemof/oemof-thermal/pull/90 on branch 'fix/rename-conf.py' 
2. rendering issue:
handled in https://github.com/oemof/oemof-thermal/pull/91 on branch 'fix/add-png-docs'

This PR is closed.

PR #84, opened 2020-03-05
Features/improve code csp

### LINKED ISSUE #79: The code for the csp is a little rough
Jann made some good suggestions to improve the code. I will adapt it soon.

Some more comments:

* in concentrating_solar_power.py:
  * The dataframe df already has a timeindex, which is internally replaced. This can lead to errors. 
  * Passing df and then telling which column means what is kind of complicated.
    I mean these 3:
    date_col: string, default: 'date'
    irradiance_col: string, default: 'E_dir_hor'
    temp_amb_col: string, default: 'temp_amb'

Suggestion: Pass irradiance (E_dir_hor or dni, depending on irradiance_method) and temp_amb as pd.Series and check that it has a valid timeindex (which is necessary for the pvlib). E_dir_hor/dni should be passed as kwargs, and depending on the method it should be checked that the right keyword argument is passed.

``` python
def csp_precalc(
                lat, long, timezone,
                collector_tilt, collector_azimuth, cleanliness,
                eta_0, c_1, c_2,
                temp_collector_inlet, temp_collector_outlet,
                temp_amb,
                a_1, a_2, a_3=0, a_4=0, a_5=0, a_6=0,
                loss_method='Janotte',
                irradiance_method='horizontal',
                **kwargs):
                """ ... """
               required_dict = {'horizontal': 'E_hor_ir', 'other': 'dni'}

               irradiance_required = required_dict[irradiance_method]

               if not irradiance_required in kwargs:
                   raise AttributeError(f"'{irradiance_required}' necessary for {irradiance_method} is not provided")

               irradiance = kwargs.get(irradiance_required)

```

* parameter names
  * tz: why not call it timezone?
  * x: why not call it cleanliness? I would be ok with x however if it is an established convention to name it like this.

* I do not quite understand the difference btw. collector_irradiance and irradiance_on_collector. The explanation in the table does not make it clear to me. Also, we should check how the pictures in the docs help understanding these (#76)

_Originally posted by @jnnr in https://github.com/oemof/oemof-thermal/pull/68#issuecomment-591354133_

FranziPl:
As mentioned in #79 the code for the csp precalculation is not in its best shape. So I changed the structur a little bit:
* There is a 'main function' which uses other function. Every single other function has an own definition.
* The input is not a dataframe enymore, but parameters to set the correct datetimeindex and series for the irradiance and the temperature
* There are more warnings, if wrong parameters are given

In this PR additionally the examples and tests should be adapted:
 [X] csp plant example is adapted
 [X] csp plot example is adapted
 [X] tests are adapted

Please let me know, if I forgot something and what you think about it in general.


FranziPl:
So I changes all the files, which are necessary.
Could one of you review the PR please?
Thanks.

c-moeller:
I cannot run the csp_collector_plot_example_andasol.py

`AttributeError: 'E_dir_hor' necessary for horizontal is not provided    `

c-moeller:
Looks good to me except for the not running example.
@jnnr Could you have a detailled look on the code since I don't know exactly what you discussed during your meeting.

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -119,26 +120,35 @@ def csp_precalc(df, periods,
...
-    else:
-        raise AttributeError("irradiance_method must be 'horizontal' or"
-                             "'normal'")
+    irradiance_required = required_dict[irradiance_method]
+
+    if irradiance_required not in kwargs:
+        raise AttributeError(
+            f"'{irradiance_required}' necessary for {irradiance_method} is not provided")
+
+    if 'temp_amb_input' not in kwargs:
```
jnnr: If temp_amb has to be provided in all cases, it would be simpler to list it among the arguments of the function together with the others:

`temp_collector_inlet, temp_collector_outlet, temp_amb`

Then you could just drop this check here.

jnnr:
Generally, I find the docstrings much easier to read if there is one empty line between Parameters. I saw that in oemof/oemof, there are no spaces as well, but still it would be an improvement I think.

But maybe that's a detail and not specific to this PR.

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -119,26 +120,35 @@ def csp_precalc(df, periods,
...
+        raise AttributeError(
+            "temp_amb_input is not provided")
 
+    date_time_index = pd.date_range(date, periods=periods,
```
jnnr: Here, a new timeindex is created.

In #79, I proposed to pass the timeseries that are input to the function as pd.Series with an timeindex. Then, the function could just check if the three series all have the same timeindex, and no new timeindex had to be created. However, as the pv-lib needs a timeindex, this would not allow the user to pass a list instead of a pd.Series. Maybe this would be good, as the user has to explicitly prepare the pd.Series with timeindex (which would be ok, a standard thing to ask), rather than letting that happening internally. What do you think?

jnnr:
I tried running the examples and got two errors:

* python csp_collector_plot_example_andasol.py 
```
Traceback (most recent call last):
  File "csp_collector_plot_example_andasol.py", line 47, in <module>
    date_col='Datum', temp_amb_col='t_amb')
  File "oemof-thermal/src/oemof/thermal/concentrating_solar_power.py", line 136, in csp_precalc
    f"'{irradiance_required}' necessary for {irradiance_method} is not provided")
AttributeError: 'E_dir_hor' necessary for horizontal is not provided
```

* python csp_plant_example.py 

```
Traceback (most recent call last):
  File "csp_plant_example.py", line 113, in <module>
    storage = solph.components.GenericStorage(
AttributeError: module 'oemof.solph' has no attribute 'components'
```

I do not fully understand the second error, but I can fix it by importing the GenericStorage directly, i.e.

`from oemof.solph.components import GenericStorage`

c-moeller:
I get the same error like @jnnr when running csp_collector_plot_example_andasol.py

but no error on csp_plant_example.py

src/oemof/thermal/concentrating_solar_power.py
FranziPl: I changed it to a positional argument and dropped the check.

src/oemof/thermal/concentrating_solar_power.py
FranziPl: I changed the structure, so now time indexed series has to be given to the function.

FranziPl:
I adapted your comments and the examples work again.

FranziPl:
I added a line to the whats new and adapted the docu.
In my opinion, it is ready to merge now.

docs/concentrating_solar_power.rst
```diff
@@ -15,26 +15,38 @@ system optimizations with oemof.solph.
...
+
+    Fig.1: The energy flows and losses at a parabolic trough collector.
+
+The direct normal radiation (E_dir) is reduced by geometrical losses
+(Q_loss,geom) so that only the collector radiation (E_coll*) hits the collector.
+Before the thermal power is absorbed by the absorber tube, also optical losses
+(Q_loss,opt), which can be reflection losses at the mirror, transmission losses
+at the cladding tube and absorption losses at the absorber tube, occur. The
+absorber finally loses a part of the absorbed heat output through thermal
+losses (Q_loss,therm).
```
c-moeller: The text somehow appears in italic letters (maybe because of the asterisk). Did you check if this happens in the compiled documentation as well? Perhaps it's only the preview which is strange here.

docs/concentrating_solar_power.rst
```diff
@@ -103,30 +115,30 @@ These arguments are used in the formulas of the function:
...
-        temp_collector_inlet, temp_collector_outlet,
-        date_col='Datum'
+        temp_collector_inlet, temp_collector_outlet, dataframe['t_amb'],
+        a_1, a_2,
+        E_dir_hor=dataframe['E_dir_hor']
         )
 
 The following figure shows the heat provided by the collector calculated with
-this function in comparison to the heat calculated with a fix efficiency.
+this functions amd the loss method "Janotte" in comparison to the heat
```
c-moeller:
```suggestion
this functions and the loss method "Janotte" in comparison to the heat
```

docs/whats_new/v0-0-2.rst
```diff
@@ -4,12 +4,14 @@ v0.0.2 (2020)
 API changes
 -----------
 
-* 
+* concentrating solar power: The input for irradiance must be a time indexed
+series now, instead of a series and separate information about the date
```
c-moeller:
```suggestion
* concentrating solar power: The input for irradiance is now a time indexed
series instead of a series and separate information about the date
```

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -17,75 +17,78 @@
...
-                temp_collector_inlet, temp_collector_outlet,
+                temp_collector_inlet, temp_collector_outlet, temp_amb,
                 a_1, a_2, a_3=0, a_4=0, a_5=0, a_6=0,
                 loss_method='Janotte',
                 irradiance_method='horizontal',
-                date_col='date', irradiance_col='E_dir_hor',
-                temp_amb_col='t_amb'):
+                **kwargs):
     r"""
     Calculates collectors efficiency and irradiance according to [1] and the
```
c-moeller:
```suggestion
    Calculates collector efficiency and irradiance according to [1] and the
```

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -17,75 +17,78 @@
...
         Thermal loss parameter 2. Required for loss method 'Janotte'.
+
     temp_collector_inlet: numeric or series with length periods
         Collectors inlet temperature.
+
     temp_collector_outlet: numeric or series with length periods
         Collectors outlet temperature.
+
+    temp_amb: time indexed series
+        Ambient temperature time series
```
c-moeller:
```suggestion
        Ambient temperature time series.
```

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -313,3 +372,27 @@ def calc_eta_c(eta_0, c_1, c_2, iam,
...
+    r"""
+
+    .. csp_heat_equation:
+
+    :math:`Q_{coll} = E_{coll} \cdot \eta_C`
+
+    Parameters
+    ----------
+    eta_c: series of numeric
+        collectors efficiency.
```
c-moeller:
```suggestion
        Collector efficiency.
```

c-moeller (changes requested):
Looks great to me. Examples are working, too.

Just two small remarks, which I think can be quickly done:
- In the example 'cps_collector_plot_example_andasol.py' you do not plot units in both figures except for the time, but style should be consistent
- In the example 'csp_collector_plot_example.py' you do not plot a unit in the second figure on ambient_temperature --> should be °C

And a question: is it Q or Q_dot? Or no difference because we have 1 hour time steps?
I think this question appeared here as well: https://github.com/oemof/oemof-thermal/pull/76

docs/concentrating_solar_power.rst
FranziPl: It looks normal in the compiled document.

FranziPl:
You are right, I added the missing units.
It is Q_dot, but Q_dot_coll doesn't seem nice to read, thats why I just wrote Q_coll. But in the units you can see, that is is power, not energy.

c-moeller (approved):

jnnr:
Shall I review this again? I can do so on Friday.

FranziPl:
@jnnr, I am not sure if it is necessary. Maybe it is a good idea for the file
\src\oemof\thermal\concentrating_solar_power.py, as I integrated a lot of your suggestions there.
The rest is fine.

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -119,77 +125,115 @@ def csp_precalc(df, periods,
...
-        raise AttributeError("irradiance_method must be 'horizontal' or"
-                             "'normal'")
+    irradiance_required = required_dict[irradiance_method]
 
-    data.set_index('date', inplace=True)
+    if irradiance_required not in kwargs:
+        raise AttributeError(
+            f"'{irradiance_required}' necessary for {irradiance_method} is not provided")
+
+    irradiance = (kwargs.get(irradiance_required))
```
jnnr: Why are there brackets? Does this have to be a tuple?

src/oemof/thermal/concentrating_solar_power.py
```diff
@@ -17,75 +17,78 @@
...
         horizontal direct irradiance or the direct normal irradiance is
         given and used for calculation.
-    date_col: string, default: 'date'
-        Describes the name of the column in the dataframe df.
-    irradiance_col: string, default: 'E_dir_hor'
-        Describes the name of the column in the dataframe df.
-    temp_amb_col: string, default: 'temp_amb'
-        Describes the name of the column in the dataframe df.
+
+    irradiance: time indexed series
```
jnnr:
```suggestion
    E_dir_hor/dni (depending on irradiance_method): time indexed series
```

src/oemof/thermal/concentrating_solar_power.py
jnnr: It seems that the brackets can be removed. Tests and examples keep running.

jnnr (approved):
Thanks for your effort! I fixed a test that did not get executed before, added another one checking for an index error and added two lines that check and raise an index error when temp_amb and irradiance do not have the same index. Also, I posted two comments, one for the docstring and one for the code. Apart from that, I am ok with merging this.

src/oemof/thermal/concentrating_solar_power.py
FranziPl: I have no idea, why they are there, I removed them

PR #87, opened 2020-03-11
Fix/whatsnew

c-moeller:
Instead of a release branch, which I rather create just before the release, I created a fix-branch for the whatsnew file. I copied the structure from the oemof repository and made some first example entries. @oemof/oemof-heat Please add entries according to your features, bug fixes and so on.


jakob-wo:
Entries for compression_heatpumps_and_chillers module are complete.

Issue #85, opened 2020-03-05
comments in gitignore file are too specific / too personalized


jakob-wo:
The gitignore file holds comments that are personalized for specific users:
```
# data from Franzi
CSP_data/
CSP_results/

# results
examples/flat_plate_collector/results/
```
I suggest either
a) to keep that structure but use git-alias to make identification more easily (Franzi --> FranziPl)
or
b) remove user names and list `CSP_data/` and `CSP_results/` under `# results`

Me personally, I prefere option b).

jnnr:
A third way would be to use a common directory name for results for all examples, and add results/ to .gitignore. 

This would replace the entry "examples/flat_plate_collector/results/"


CSP_data seems to be orphaned and can be deleted from .gitignore. @FranziPl, what do you think?

PR #92, opened 2020-03-25
Remove personal entry from gitignore

FranziPl:
I removed my personal entries from the gitignore file to solve #85. I added my directory to results, because I still need it.
@jnnr I am not sure, how we can bring the results together in one entry, because they are in different directories?


jakob-wo:
I am not sure but maybe something like _examples/**/results_ in the gitignore-file will work as a generic solution.
Therefore it would be necessary to rename _csp_results_ --> _results_.

FranziPl:
So I changed the name of the folder, so one entry for all results should work.
@c-moeller and @jakob-wo, can you have a look, if it works for you too?

jnnr:
I agree with the changes made here. Two things have to be fixed however before merging. For the examples in concentrating_solar_power and solar_thermal_collector, the results directory has to be created before the results are stored. Use `os.mkdir()`. Otherwise, the user gets a "path not found" error.

jnnr (changes requested):
I agree with the changes made here. Two things have to be fixed however before merging. For the examples in concentrating_solar_power and solar_thermal_collector, the results directory has to be created before the results are stored. Use `os.mkdir()`. Otherwise, the user gets a "path not found" error.

```
results_path = os.path.join(path,'examples', 'solar_thermal_collector', 'results')
if not os.path.exists(results_path):
    os.mkdir(results_path)
```

Adapted the code accordingly (see the commits below).

jnnr (approved):

FranziPl:
Thanks @jnnr for creating the paths.

PR #94, opened 2020-04-02
Features/facade solar thermal collector

MaGering:
This PR implements a facade for the solar thermal collector. Related issue: #35
It is based on the facade built for the csp component: https://github.com/oemof/oemof-thermal/pull/77

It so far contains drafts of:
- a class for the solar thermal collector in facades.py
- an example of a facade based on flat_plate_collector_example.py

In order to work with the approach of comparing lp-files (as described in https://github.com/oemof/oemof-thermal/pull/43) 
- flat_plate_collector_example.py has been updated with the generation of a lp-file 
- test_constraints.py has been added to tests


examples/solar_thermal_collector/flat_plate_collector_example_facade.py
```diff
@@ -0,0 +1,140 @@
...
+conversion_storage = 0.98
+size_collector = 1000 # m2
+
+# busses
+bth = solph.Bus(label='thermal')
+bel = solph.Bus(label='electricity')
+
+collector = facades.SolarThermalCollector(
+    label='solar_collector',
+    output_bus=bth,
```
jnnr: For clarity:
heat_out_bus,
electricity_in_bus
Should be consistent with CSP.

examples/solar_thermal_collector/flat_plate_collector_example_facade.py
```diff
@@ -0,0 +1,140 @@
...
+    electrical_bus=bel,
+    electrical_consumption=elec_consumption,
```
jnnr, line 74: What is the meaning of electrical consumption? What are the units? I guess this is something like electricity input/heat output. Could it be defined as a (pump) efficiency? The docs do not mention this parameter yet.

examples/solar_thermal_collector/flat_plate_collector_example_facade.py
```diff
@@ -0,0 +1,140 @@
...
+    peripheral_losses=eta_losses,
+    aperture_area=size_collector,
+    latitude=latitude,
+    longitude=longitude,
```
jnnr, line 78:
```suggestion
    lat=latitude,
    lon=longitude,
```

jnnr: Unify with CSP.

examples/solar_thermal_collector/flat_plate_collector_example_facade.py
c-moeller: I'm not sure if I did it as you intended. In the CSP facade the bus names are sligthly different.

examples/solar_thermal_collector/flat_plate_collector_example_facade.py
c-moeller: Hmm... am I on the wrong branch? On features/facace_csp it's latitude and longitude..

jnnr:
Running the example, I get the same error as described in #77. When using pandas 0.24.2, you have to set the attribute freq='H' of the timeindex, otherwise you get an error back when creating the oemof Model.

```
AttributeError: No valid time increment found. 
Please pass a valid timeincremet parameter or pass an EnergySystem with a valid time index.
Please note that a valid timeindex need to have a 'freq' attribute.
```

MaGering:
I've implemented the saving of the model results in a csv file for the facade's example: https://github.com/oemof/oemof-thermal/pull/94/commits/49b90d3a8679272461b2d71bae8dbd12b724ca74

Unfortunately the results of the examples and the facade examples model do not match yet.

c-moeller:
Almost ready to merge. Two things are open.

@FranziPl Can you have a look at my results output? I still seem to have an error with the efficiency parameters. Electricity from the grid, excess of the collector and the collector's heat are different in the facade and in the non-facade example. I'm lost :-(

@jnnr I seem to have the same failed test like the CSP facade. Can you adapt this like you did it in the CSP facade or tell me what to do?

FranziPl:
The collector is far too big for the demand. That's why most of the heat is going to the excess. In the example without facade, it is going to the excess from the solar bus, in the facade example it is going to excess from the thermal but, because there is no solar bus. That is why the heat from the collector to the thermal bus is not the same.
Change the size of the collector from 1000 to 10 m2 and you can compare the results better. Than the flows (('collector', 'thermal'), 'flow') and (('solar_collector', 'thermal'), 'flow') are the same. Also electricity and storage flows are the same.

FranziPl:
I am not sure, if the plots make sense for the examples. Is it informative to show the difference of precalculation and a constant efficiency here? This is independent from the facade and the example.

FranziPl:
In file "solar_thermal_collector.py" the docstring is not adapted to the new parameters (line 63 there are parameters, which don't exist anymore: date_col, irradiance_global_col, irradiance_diffuse_col, temp_amb_col: string).
As it wasn't changed in this PR I couldn't mark it in the "Files changed" tab.

Issue #96, opened 2020-04-03
Missing scheme of solar thermal collector in docs


jnnr:
In #76, a scheme of the solar thermal collector has been added. However, the image was not included in docs/solar_thermal_collector.rst, so it is not shown. In a new PR, the image should be included.

PR #103, opened 2020-04-09
Show schematic of flat plat collector in docs

jnnr:
Addresses #96.


MaGering (approved):
Thank you for this PR! I've added missing dots over Q. I guess it can be merged now.

PR #107, opened 2020-04-20
Fixes/docstring thermal storage

jnnr:
The docstring of the stratified thermal storage facade has not been updated. This PR fixes this.

### DIFF
--- a/src/oemof/thermal/facades.py
+++ b/src/oemof/thermal/facades.py
@@ -133,6 +133,8 @@ class StratifiedThermalStorage(GenericStorage, Facade):
 
     Parameters
     ----------
+    label : str
+        Unique label.
     bus: oemof.solph.Bus
         An oemof bus instance where the storage unit is connected to.
     diameter : numeric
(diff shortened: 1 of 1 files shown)


c-moeller (changes requested):
Still miss some attributes in the docstring:
self.height
self.water_properties (heat_capacity, density)
self.minimum_storage_capacity
self.marginal_cost

And I cannot see the following explained attributes to be part of the class:
initial_storage_level
min_storage_level
max_storage_level

jnnr:
Quoting from https://www.python.org/dev/peps/pep-0257/

> If a class subclasses another class and its behavior is mostly inherited from that class, its docstring should mention this and summarize the differences. Use the verb "override" to indicate that a subclass method replaces a superclass method and does not call the superclass method; use the verb "extend" to indicate that a subclass method calls the superclass method (in addition to its own behavior).

So the docstring of a facade like StratifiedThermalStorage should summarize all attributes that come in addition to the attributes of the parent classes (Facade and GenericStorage in this case).

src/oemof/thermal/facades.py
```diff
@@ -133,10 +133,28 @@ class StratifiedThermalStorage(GenericStorage, Facade):
...
+    temp_env : numeric
+        Temperature of the environment.
+    u_value : numeric
+        Thermal transmittance [W/(m2*K)]
+    initial_storage_level : numeric
+        Storage level at the start (between 0 and 1).
+    min_storage_level : numeric
+        Minimal storage level (between 0 and 1).
+    max_storage_level : numeric
+            Maximal storage level (between 0 and 1).
```
c-moeller:
```suggestion
        Maximal storage level (between 0 and 1).
```

c-moeller (approved):

PR #113, opened 2020-04-28
Fix/hotfixes

c-moeller:
Hotfixes for v0.0.2


jnnr:
I tested the examples and they run. Some remarks:
* The plots of the two examples are stored under the same name, but the facade example creates another plot 'heat of the collector'. Also, the plot 'heat of the collector' is shown and saved, the others only saved. Maybe this could be clearer? E.g. by saving the plots and `print`ing a message that says 'Plot a and b have been saved in /results.'
* The extra plot in the facade example does not use the same function as the other plots (which is imported from plots.py). plots.py is not an example itself.
* The plots do not have a legend.

examples/solar_thermal_collector/plots.py
```diff
@@ -27,7 +27,7 @@ def plot_collector_heat(data_precalc, periods, eta_0):
     )
     ax.grid()
     ax.legend()
-    plt.savefig('compare_precalculations.png')
+    plt.savefig('results/compare_precalculations.png')
     plt.show()
```
jnnr, line 31:
```suggestion
```

jnnr (commented):
I tested the examples and they run. Some remarks:
* ~~The plots of the two examples are stored under the same name, but the facade example creates another plot 'heat of the collector'. Also, the plot 'heat of the collector' is shown and saved, the others only saved. Maybe this could be clearer? E.g. by saving the plots and `print`ing a message that says 'Plot a and b have been saved in /results.'. Or, not saving at all.~~
**Update**: Both examples create different plots ('compare_precalculations.png' 'heat_of_the_collector.png'). Both of them are shown and saved. One of them imports a plot function, the other has the plot function included in the script. 
* The extra plot in the facade example does not use the same function as the other plots (which is imported from plots.py). plots.py is not an example itself.
* The plot 'heat of the collector' does not have a legend.

c-moeller:
I fixed the investment example and added it again.

@jnnr Thanks for your remarks. I do not understand the first point. Both plots are saved under a different name and shown while running. I do not see the point. Second point is a problem, I know.. but I would like to shift it to v0.0.3. Last point: the legend is not shown since there is only one line.

jnnr:
> @jnnr Thanks for your remarks. I do not understand the first point. Both plots are saved under a different name and shown while running. I do not see the point.

Seems that I have got it wrong. Both examples create different plots ('compare_precalculations.png' 'heat_of_the_collector.png'). Both of them are shown and saved. One of them imports a plot function, the other has the plot function included in the script. 

> Last point: the legend is not shown since there is only one line.

In that case, you could just remove `ax.legend()`.

Issue #120, opened 2020-05-26
Missing documentation of theoretical considerations regarding aggregation


MaGering:
The issue https://github.com/oemof/oemof-solph/issues/647 describes the approach of the aggregation of domestic decentral solar thermal systems. The documentation of this approach is still missing in oemof.thermal.

PR #121, opened 2020-05-26
Features/docs aggregation

MaGering:
With this PR the documentation of the theoretical considerations towards the aggregation of domestic decentral solar thermal systems is created.

Related issues: https://github.com/oemof/oemof-thermal/issues/120 and https://github.com/oemof/oemof-solph/issues/647

The structure of the section in the documentation is based on the content of issue https://github.com/oemof/oemof-solph/issues/647.


docs/theoretical_considerations.rst
```diff
@@ -0,0 +1,38 @@
...
+
+The system shown in the picture can be characterized by two ratios:
+
+:math:`ratio_{1} = \frac{V_{storage}}{A_{collector}} \quad \textrm{and} \quad ratio_{2} = \frac{A_{collector}}{\dot{Q}_{demand,max}}`
+
+or, also possible:
+
+:math:`ratio_{1} = \frac{V_{storage}}{A_{collector}} \quad \textrm{and} \quad ratio_{2} = \frac{A_{collector}}{\dot{Q}_{total}}`
+
+Usually these ratios are constant within the two types.
```
c-moeller:
```suggestion
Normally there are typical ratios which are used when dimensioning these types of systems.
```

docs/theoretical_considerations.rst
```diff
@@ -0,0 +1,38 @@
...
+So it is possible to aggregate the houses within one type and define the ratios. This would be more accurate than an aggregation of all installed systems.
```
c-moeller:
```suggestion
So it is possible to aggregate the houses within one type by defining these ratios. This would be more accurate than an aggregation of all installed systems.
```

docs/theoretical_considerations.rst
```diff
@@ -0,0 +1,38 @@
...
+If there are special system modifications, there can be defined more than two groups.
```
c-moeller:
```suggestion
In case of special system modifications, also more than two groups could be defined.
```

docs/theoretical_considerations.rst
```diff
@@ -0,0 +1,38 @@
...
+.. 	figure:: _pics/aggregation_scheme.png
+   :width: 70 %
+   :alt: aggregation_scheme.png
+   :align: center
+
+Concept
+_______
+
+
+Instead of aggregate all storages and all collectors of the systems, which would be really unaccurate, we want to classify two different types of domestic solar systems:
```
c-moeller:
```suggestion
Instead of aggregate all storages and all collectors of the system, which would be really unaccurate, we want to classify two different types of domestic solar systems:
```

c-moeller (changes requested):
I added some small changes.

And one request: Could you put the section "Theoretical considerations" on the same level as "User's Guide". And "Aggregation..." on the same level as the components in "User's Guide"

And one more remark: We should have some kind of nomenclature, even though the symbols seem self-explanatory. So, if it is not too much effort, you could add a small table (but I would say with lower priority).

MaGering:
> And one more remark: We should have some kind of nomenclature, even though the symbols seem self-explanatory. So, if it is not too much effort, you could add a small table (but I would say with lower priority).

I've added a first draft of a nomenclature in the section 'Aggregation...'. Could you please check it and round it off @c-moeller ?

c-moeller (approved):

PR #125, opened 2020-06-05
Update to oemof.solph v0.4

### LINKED ISSUE #124: Metapackage named as dependency
With the release of solph v0.4, the result should be oemof.solph instead of oemof.

p-snft:
Fixes #124


jnnr:
Thanks for this PR! Looks good. However, the examples have to be adapted as well.

jnnr (changes requested):
Thanks for this PR! Looks good. The tests run successfully at my local pc. However the examples do not work. It would be best to fix them as well. And there are two constraint tests of the stratified thermal storage that have been ignored until now but can be included as well.

**Update**: I fixed the last point myself quickly, adding the tests and updating the lp-files.

MaGering:
I've updated the examples to oemof.solph v0.4. Some example are still not running however because of oemof site-packages (e.g. oemof.outputlib.processing).

@c-moeller could you take a look at this?

c-moeller:
Thanks @MaGering!
The imports don't work for me either. But I think the problem is because of the outputlib. This seems now part of oemof-solph and does not need to be imported separately.

c-moeller:
But I still have some import errors. I cannot import sources, busses and so on from oemof.solph...

c-moeller:
@jnnr Could you have a look at the last commit. The stratified thermal storage examples are not running. We only removed the 'outputlib' and imported the modules 'processing' and 'views' directly from solph which was sufficient for every other example. But somehow 'yours' are not working.

In your virtualenv you need to
- install oemof-thermal
- uninstall oemof
- install oemof-solph

jnnr:
I adapted the stratified thermal storage examples, now they run with oemof 0.4.1.

jnnr:
Both solar thermal collector and concentrating solar collector throw a SuspiciousUsageWarning: 

SuspiciousUsageWarning: `Transformer` 'solar_collector' constructed without `outputs`.
  warn(msg.format(self), debugging.SuspiciousUsageWarning)

FranziPl:
In which file or which situation?
For csp, I only found the name 'solar_collector' in the example_csp_facade.py file. There is a component, which is named like that and uses the csp_facade. It does create an output.
So I assume, that the problem is about the facade for csp and flate plate collector (as they have a similar structure).
In the facade in the function 'build_solph_components' the output is defined with the method 'self.outputs.update'. I am wondering, if this method could be a problem, but I copied it from oemof tabular.
Summary: I have no clue.

c-moeller (approved):

c-moeller:
I will shift the SuspiciousUsageWarning subject to the next release.

Issue #128, opened 2020-06-16
API facades.StratifiedThermalStorage does not include all parameters in list


jakob-wo:
The parameter list of the facades.StratifiedThermalStorage API misses the parameters:
`label`, `initial_storage_level`, `max_storage_level`, `min_storage_level` and `invest_relation_input_capacity`, `minimum_storage_capacity`.

jakob-wo:
What about the `nominal_storage_capacity`, should that appear in the parameter list?
I am not sure how that variable is handled. Btw.: Is ist possible to use `nominal_storage_capacity` as a direct input parameter without calculating it from the dimensions?

c-moeller:
As far as I understood, these are parameters of the GenericStorage and therefore not described again.

jakob-wo:
Well, now that you mention it it's obvious :) 
I did not realize it before for some reason...

Perhaps it would help to point out in the documentation that the facade bases on the GenericStorage. Until now the docu does not mention that but one have to look up the API and find the tiny note that facade bases on GenericStorage.

jnnr:
I suggest that you come up with a text that would have helped you and makes things clearer. You can assign me as reviewer in the PR.

jakob-wo:
I have another question about the arguments that can be passt to the facade...
Do I understand you right, that all arguments of GenericStorage can also be passt to the facade? What happens then if I, for example, add a value for `nominal_storage_capacity` to the facade in the example `examples/stratified_thermal_storage/stratified_thermal_storage_facade.py`? 
Does the facade calculate the nominal storage capacity from the dimensions and temperatures or does it take the value I passed directly?

jakob-wo:
@jnnr, I think it would still be interesting to clearify the handling/priorisation of parameters within the facade that can be passed directly and/or via precalculation (question/example above).

jnnr:
To conclude: We can make it easier for users if we add 2 things to the docs:

* Explain in a general place in the text that facades inherit from the generic component classes of oemof.solph. In most cases, the attributes of the parent class can be passed to the facade as well.

* Explain in the docstring of StratifiedThermalStorage: The attribute :attr:`nominal_storage_capacity` should not be passed because it is determined internally from :attr:`height` and :attr:`parameter.`

jnnr:
The above suggestions have been implemented and merged with #165

PR #134, opened 2020-06-17
Improve documentation of stratified thermal storage

jakob-wo:
The aim of the PR is to clearify in the docu text that the facade bases on GenericStorage and that all of its parameters can therefore also be used with the facade of the stratified thermal storage.

Addionally, the information could be added to the API if that appears usefull.


jakob-wo:
This PR is realted to Issue #128.

Issue #140, opened 2020-06-22
Instruction on how to install developer version is missing in README


MaGering:
In the READMEs of oemof.solphs and of oemof.network there is an instruction on how to install the developer version of the repositories.

Would it be a nice to have for oemof.thermal too? And can it be installed the same way (pip install https://github.com/oemof/oemof-thermal/archive/dev.zip) as done in the above mentioned repositories?

PR #141, opened 2020-06-22
Add explanation how to install developer version to README

MaGering:
This PR adds an instruction on how to install the developer version of oemof.thermal.

Further information can be found in the related issue https://github.com/oemof/oemof-thermal/issues/140


c-moeller (commented):
I think under 'Contributing' it has been already sufficiently described, because it leads to the general developing rules of oemof

Issue #127, opened 2020-06-15
Import oemof.tabular's base class Facade


jnnr:
Currently, the facades in oemof.thermal inherit from a base class that has been copied from oemof.tabular. This was done in the beginning to try the concept of facades and reduce the number of imports. Now, as both packages, oemof.tabular and oemof.tabular, get used more and more, we can as well import oemof.tabular. This would help to avoid divergent developments, keep compatibilty and would be in line with the new unbundled repo structure.

c-moeller:
I would suggest to shift this to v0.0.4 due to the workload in v0.0.3 and current activity in oemof-tabular. Is this ok for you?

jnnr:
It could be done within a couple of minutes. This is not a lot of work in terms of programming. It may generate some work in the future, if changes have to be coordinated with oemof.tabular. And it will also keep things easier when someone wants to use the two packages together. 

It is a decision. if we need some more time for that, there is no need to hurry.

c-moeller:
Ok, I set it back to v0.0.3

jnnr:
Currently, importing from oemof.tabular would induce a version conflict:

* oemof.thermal relies on oemof.solph v0.4.0 or higher,
* oemof.tabular is still not updated to oemof.solph v0.4.0.

Therefore, we should not import from oemof.tabular. Once oemof.tabular is updated, we could decide to import from there. I removed the milestone. Shall I leave this issue open as a reminder for later?

c-moeller:
I would leave it open.

PR #142, opened 2020-06-23
Add import of Facade class from oemof tabular's facades module

MaGering:
This PR
- replaces the copy of the oemof.tabular.facades module in oemof.thermal's facades.py with an import of the required class (Facade) and
- deletes obsolete imports

Related: Issue https://github.com/oemof/oemof-thermal/issues/127


c-moeller:
Due to incompatibilities of package requirements when installing oemof-tabular together with oemof-solph v0.4.1, I would like to shift this to v0.0.4 If someone gets this running until tomorrow noon, feel free, otherwise let's shift this.

Issue #101, opened 2020-04-08
examples and docu of csp doesn't explain important details


FranziPl:
As mentioned in PR #77, there are some improvements, which should be done, but which are not part of the PR:
* [x] mention in docs that time index has to be in the time zone consistent to the position defined by lat/long
* [x] In example: Here we define the timezone dataframe.index.tz_localize 
* [x] leave out unused data from input data
* [x] pandas problems freq='H' with version 0.24.2
* [x] give warnings, if the selected methods don't match the number of parameters

FranziPl:
I changed the points 1 to five in PR #147.

FranziPl:
I shift the 5th point to v0.0.4

FranziPl:
Last Point is solved in PR #164.

PR #147, opened 2020-07-01
Features/improve example explanation csp

FranziPl:
This PR will adress the explanation problems from #101.
It will change:

- [x] mention in docs that time index has to be in the time zone consistent to the position defined by lat/long
- [x] In example: Describe, where the timezone dataframe.index.tz_localize is defined.
- [x] leave out unused data from input data
- [x] mention pandas problemswith version 0.24.2, so freq='H' is needed


FranziPl:
@c-moeller : I assigned you, because of a needed review. As this PR just add information in comments, no functionality, it is fine, if you don't want to do a review. Midday I will check and merge the PR.

c-moeller (approved):
Looks good!

PR #150, opened 2020-07-02
Use collector heat as max instead of fix

p-snft:
The solar thermal façades set their power gains as a fixed quantity.
This means, all of it has to go somewhere. It would prevents energy
systems from being unsolvable because of heat overproduction,
when it is seen as the maximum instead.


jnnr:
This sounds reasonable to me. But maybe I am overlooking something. What do you think, @FranziPl?

c-moeller:
In the non facade example we combine the collector with a sink in case of excess heat, but I cannot find this in the facade either. But maybe I'm also overlooking something.

FranziPl:
I think it is a good idea. At present I defined an excess sink also in the facade example. It seems to be a better option.

FranziPl (approved):

PR #161, opened 2020-09-09
Fixes/docs whatsnew

### LINKED ISSUE #160: What's New information on v0.0.3 release is missing in documentation
https://oemof-thermal.readthedocs.io/en/latest/whats_new.html

I have no idea why, since the file exists: https://github.com/oemof/oemof-thermal/blob/dev/docs/whats_new/v0-0.3.rst

@oemof/oemof-heat Can somebody help?

jnnr:
Fixes #160.


c-moeller:
Great thank you! What happened to the tests?

jnnr:
The csp test fails. Didn't that happen before? Do we have a solution for that?

**update**: It fails at my computer locally as well. The difference is in the last decimal digit. Maybe some tolerance when comparing lp-files would help.

PR #164, opened 2020-09-25
Add a warning, if required arguments are missing

FranziPl:
This PR should solve Issue #101. There will be a warning, if required parameters are missing. It is a warning and not an error, if somebody wants to set parameters to 0 on purpose.


FranziPl:
The coverage decreases, but it is just the line for the warning (I have no idea how to test warnings, I can just test errors).

c-moeller (approved):

PR #166, opened 2020-09-28
Features/csp validation

FranziPl:
This PR adds the code for the validation of the csp component. To publish the data, we need an approval first.


examples/concentrating_solar_power/validation.py
```diff
@@ -0,0 +1,151 @@
+# -*- coding: utf-8 -*-
+
+"""
+Date: 17.09.2020
+Author: Franziska Pleissner
+
+This file shows a comparison between the collectors heat, which is calculated
```
c-moeller:
```suggestion
This file shows a comparison between the collector heat, which is calculated
```

examples/concentrating_solar_power/validation.py
```diff
@@ -0,0 +1,151 @@
...
+with the oemof.thermal component and the collector heat, which is measured.
```
c-moeller:
```suggestion
with the oemof-thermal component and the collector heat, which is measured.
```

examples/concentrating_solar_power/validation.py
```diff
@@ -0,0 +1,151 @@
...
+It uses the irradiance data, which are measured at the same spot and the same
+time as the collector heat data
```
c-moeller:
```suggestion
time as the collector heat data.
```

c-moeller (changes requested):
I wonder if we better wait for the approval of the data, so that the file is running. Do you know, if we get this approval for sure and it's just a matter of time?

FranziPl:
I am not sure, but optimistic.

c-moeller:
Ok, then let's wait with the merge.

PR #168, opened 2020-09-29
Examples: Stratified thermal storage clean-up

jnnr:
* Renaming the examples for clarity
* Adding explanation as a comment on top of each example
* Make sure that input data is taken from data/stratified_thermal_storage.csv

Open TODOs

* [x] first example 01_....py plot has changed/looks like nothing is happening. **Update:** Seems to have been a problem with my oemof version, which was not up-to-date.
* [x] model_validation.py fails: oemof.solph has no attribute 'views'. Is this a local problem for my installation? Seems to be consistent with oemof.solph v0.4.1 **Update:** Same as above, a problem with my oemof version, which was not up-to-date.


examples/stratified_thermal_storage/03_investment_fixed_ratio_generic_storage.py
```diff
@@ -1,7 +1,8 @@
 """
-For this example to work as intended, please use oemof-solph v0.4.0 or higher
-to ensure that the GenericStorage has the attributes
-`fixed_losses_absolute` and `fixed_losses_relative`.
+This example shows how to invest into nominal_storage_capacity and capacity
+(charging/discharging power) with a fixed ratio. Pass invest_relation_input_capacity
```
c-moeller:
```suggestion
(charging/discharging power) with a fixed ratio. Pass invest_relation_input_capacity and
```

c-moeller (approved):

PR #175, opened 2021-01-30
Remove upper dep limit for pandas and numpy

sasanjac:
The upper limit on pandas and numpy leads to dependency issues when installing other oemof components such as demandlib which requires pandas >= 1.0 and numpy >= 1.17.0.


jnnr:
Hi there! Thanks for opening a PR. The upper limits have been introduced with https://github.com/oemof/oemof-thermal/commit/edfd565aa489be4d3d1f8fcdfc4e4f9dc427a473 by @c-moeller, but unless there is a good reason for them still, it makes sense for them to be removed. Do you have any objections against removing them, @c-moeller?

Your commit seems to have introduced some other changes that do not seem related or intended. Would be best to open a new branch and PR with a clean commit.

Thanks for your commitment!

sasanjac:
Ok, thanks for the feedback. I removed the additional changes and will create another PR with them.

jnnr:
Thanks for fixing that. I thought you might want to start with a clean branch. But it is ok to revert. You don't have to open a new PR.

c-moeller:
I don't remember exactly, but I think I introduced these limits, because without those I had dependency issues, too. Where do the depencency limits of demandlib and oemof-thermal interfere? It seems to me, that the demandlib dependencies are still fulfilled :thinking:

sasanjac:
demandlib requires `pandas >= 1.0` and oemof-thermal currently requires `pandas >= 0.18.0, < 0.26`.
Usually, upper dependency limits are only set, if the dependency introduced breaking API changes which occurred in pandas 1.0 ([https://dev.pandas.io/docs/whatsnew/v1.0.0.html#backwards-incompatible-api-changes](url)) but they didn't have any effect in regards to the tests. I don't know if you added these limits because there actually were changes which affected oemof-thermal and were not caught in the tests. If this is the case, I can open another PR and make some suggestions for more resilient testing, if it is desired.

p-snft:
Personally, I use a local copy of thermal that allows pandas => 1.x without any problems (and in combination with demandlib). So, I am strongly in favour of removing the upper limit(s).

PS: I introduced the lower limit in demandlib. (The previous version was using a bug in pandas, that was recently fixed. So I had to decide for one version – and I chose the one without the bug.)

c-moeller:
> demandlib requires pandas >= 1.0 and oemof-thermal currently requires pandas >= 0.18.0, < 0.26

Ah ok, now I see the conflict. If it works, then I have nothing against removing the limits. As I said, I cannot really remember the reasons of the limits. Thank you for your work!

p-snft:
The tests seem to fail because of errors that were fixed in dev. I will now merge as I desperately need this.

Update: The bugs were fixed in dev.

jnnr (approved):
Looks reasonable. Haven't tested it though to find out if this introduces any problems. If it does not, I approve.

PR #178, opened 2021-02-11
Name submodules in init

p-snft:
For better compatibility, submodules shoud be named in the init.
Otherwise, the following is not possible:
```Python
    from oemof import thermal
    thermal.stratified_thermal_storage.calculate_losses(...)
```

### DIFF
--- a/src/oemof/thermal/__init__.py
+++ b/src/oemof/thermal/__init__.py
@@ -1,11 +1,10 @@
 __version__ = '0.0.5dev'
 __project__ = 'oemof.thermal'
 
-from . import (absorption_heatpumps_and_chillers,
-               compression_heatpumps_and_chillers,
-               facades,
-               stratified_thermal_storage,
-               cogeneration,
-               concentrating_solar_power,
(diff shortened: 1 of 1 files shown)


p-snft:
Without this, I get
```
AttributeError: module 'oemof.thermal' has no attribute 'stratified_thermal_storage'.
```
on some computers. The problem is comparable to the one in https://github.com/oemof/oemof-examples/issues/69, which was solved by https://github.com/oemof/oemof-tools/pull/6 in the same way I suggest here.

jnnr (approved):
Thanks for opening this. Sounds reasonable. And should not have any implied changes for usage, except for fixing the bug of course. I can test it, if you want, I guess early next week. If you don't need it, proceed.

p-snft:
> Thanks for opening this. Sounds reasonable. And should not have any implied changes for usage, except for fixing the bug of course. I can test it, if you want, I guess early next week. If you don't need it, proceed.

It was rather urgent to have this working at least in the dev version. Thank you!

PR #188, opened 2024-11-27
Release v0.0.7

p-snft:
This is a maintenance release to allow continued usage of oemof.thermal.


p-snft:
Only reporting to Coveralls fails.

